PR #44, opened 2017-10-25
Added short flag syntax for command options

calebkleveter:
This adds an optional property to the `Option` type, so it can be initialized with a short flag syntax, along with the full name:

```swift
Option(name: "foo", short: "f", help: [
    "Nottin' here."
])

Option(name: "another", help: [
    "HELP! HELP!"
 ])
```

When a short flag is added to the option, it is output in the help message:

```
Usage: ./.build/x86_64-apple-macosx10.10/debug/Executable fake [--foo -f] [--bizbaz -b] [--another]


Arugments:

Options:
    foo f Nottin' here.
 bizbaz b Here is some advice
          Don't use this command
  another HELP! HELP!
```

Short flags are also supported in the `Sequence.options` computed property, so they can be accessed the same way as long flags:

```
./.build/x86_64-apple-macosx10.10/debug/Executable fake -f -b=smile --another=one

// arguments.options = ["another": "one", "f": "true", "b": "smile"]
```

### DIFF
--- a/Sources/Console/Command/Command+Print.swift
+++ b/Sources/Console/Command/Command+Print.swift
@@ -52,15 +52,16 @@ extension Command {
         
         console.info("Options:")
         for opt in opts {
-            let shortLength = opt.short != nil ? 2 : 0
+            let shortLength = opt.short != nil ? 3 : 0
+            let longFlagDashLength = 2
             console.print(String(
-                repeating: " ", count: width - opt.name.characters.count - shortLength),
+                repeating: " ", count: width - opt.name.characters.count - longFlagDashLength - shortLength),
                 newLine: false
             )
             if let short = opt.short {
-                console.success("\(opt.name) \(short)", newLine: false)
+                console.success("-\(short) --\(opt.name)", newLine: false)
             } else {
-                console.success(opt.name, newLine: false)
+                console.success("--\(opt.name)", newLine: false)
             }
             
             for (i, help) in opt.help.enumerated() {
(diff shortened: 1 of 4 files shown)


joscdk:
I really like the idea of short flags available.

Maybe we should even change the help output to add `-` and `--` for options, take a look how Docker does it..

screen shot 2017-10-25 at 11 07 20 am

tanner0101:
@calebkleveter can you make it so the value in the `arguments.options` dictionary is the long name? (not 100% sure if that's already happening or not).

```swift
Option(name: "foo", short: "f", help: [
    "Nottin' here."
])

Option(name: "another", help: [
    "HELP! HELP!"
 ])
```

```sh
vapor -f=hi --another=sup
```

```swift
print(arguments.options) // ["foo": "hi", "another": "sup"]
```

calebkleveter:
@joscdk Implemented! Working on @tanner0101 request now.

Joannis:
Good idea @tanner0101 , that's really solid thinking

calebkleveter:
Alright, implemented the suggestion from @tanner0101. I also fixed the usage signature length calculation (which I incorrectly calculated before 😊).

Sources/Console/Command/Command+Print.swift
```diff
@@ -15,9 +15,7 @@ extension Command {
     }
 
     public func printSignatureHelp() {
-        let maxWidth = getSignitureLength()
-        let leadingSpace = 2
-        let width = maxWidth + leadingSpace
+        let width = getSignitureLength(withPadding: 2)
```
vzsg: "Signature" has a typo here and where the function is declared below.

calebkleveter: Fixed.

calebkleveter:
@tanner0101 Is this ready to be merged?

PR #74, opened 2018-09-07
Added Optional Arguments to Command

calebkleveter:
This PR adds an `optional` property to the `CommandArgument` struct:

```swift
CommandArgument.optional(name: "foo", help: ["Argument not required"])
```

When a command's arguments are parsed from the input, no error will be thrown if the argument is not found. You can optionally access the argument using the `context.arguments` dictionary property:

```swift
if let foo = context.arguments["foo"] {
    // Code here...
}
```


calebkleveter:
@tanner0101 Good points. For now, I can't think of a way to eliminate this unless we did keyed arguments:

```
my-command foo:hello bar:world
```

I don't think we want this, but maybe it would be fine?

I'll close this PR for now and we can reopen if we come up with a solution.

Issue #76, opened 2018-10-01
Command Options do not pick up default values.


EricWVGG:
sample project: https://github.com/EricWVGG/vapor-command-problem

This command should respect "one" as the default value for "answer"
`vapor run quiz` 
Expected response: `better not be the middle one`
Actual response: `What?`

EricWVGG:
Thank you for addressing my other issue. Regarding this one, this still strikes me as broken. Is it filed in the correct repo?

vi4m:
It works when you procide empty "option"

` vapor run quiz --answer`

vi4m:
Ok, I re-read the documentation, and it seems, it's really a bug here. Have fix already :)

vi4m:
Fixed here: https://github.com/vapor/console/pull/79

PR #79, opened 2018-10-19
Fix default option for commands #76

vi4m:


Sources/Command/Run/CommandContext.swift
```diff
@@ -71,6 +71,10 @@ public struct CommandContext {
 
         for opt in runnable.options {
             parsedOptions[opt.name] = try input.parse(option: opt)
+            
+            if case let CommandOptionType.value(defaultValue) = opt.type, defaultValue != nil  {
+                parsedOptions[opt.name] = defaultValue
```
calebkleveter: Will this overwrite the parsed value with the default value if a default value is found?

vi4m: Thanks for catching this. I provided test for this, since there was none.

vi4m:
Done. Anything else?

ldstreet:
If this is working as expected, maybe clearer documentation is in order. I was confused as well and expected the default value to come through when the argument was provided. I guess it depends on the most common use case. For instance, in my situation I have a `--destination` option which needs to have a default of "." and doesn't really make sense for it to be nil. I'm struggling at the moment to think of an option that could be nil, but I am sure there are some. I do think that it would be nice to work with the option with it being optional. Perhaps there is a use case for `CommandOptionType ` having the cases `.defaultedValue(default: String)`, `.value`, and `.flag`? That way when access `CommandContext` an option with type `.defaultedValue` can return a non optional value which defaults to `default`, .value can return an optional value or non optional with `requireOption`, and `.flag` can return a boolean.

vi4m:
Not sure about the reasonable default here, so I've searched for other "prior art" at my second programming language, being Python. 

I've found  Python's implementation of "OptionParser" defaults to `not nil` when default is defined by the programmer.

Example:

zrzut ekranu 2019-01-14 o 20 28 52

zrzut ekranu 2019-01-14 o 20 27 21

calebkleveter:
I think adding a flat to the `CommanOption.value` case so this can be customized might be better. So if you have this:

```swift
CommandOption.value(name: "number", short: "n", default: "0")
```

Not passing in the `-n` flag gives `nil`, but if you have this:

```swift
CommandOption.value(name: "number", short: "n", default: "0", defaultInAbsence: true)
```

You get `"0"`

tanner0101:
Maybe we can rename `default` to something else? Maybe `flagValue:`?

PR #81, opened 2018-10-31
only output help for command errors

LoganWright:
It's getting pretty bogged down in the help command when it outputs for every error. Having just CommandError logs when uses command incorrectly, but not when command fails. That way the failure output can direct them more clearly


penny-coin:
Hey @LoganWright, you just merged a pull request, have a coin! 

You now have 507 coins.

PR #93, opened 2019-03-22
Change "value option" matching to require an exact match for its "short" variant

MrMage:
Motivation: `Environment.detect` calls `try commandInput.parse(option: .value(name: "env", short: "e"))`. We would not want to match e.g. `-NSTreatUnknownArgumentsAsOpen NO` (passed by Xcode to some tests) be interpreted in the same way as `--environment NO` would. Instead, only e.g. `-e test` should be matched.


MrMage:
> Nice! Would you mind adding a test for this?

Done.

MrMage:
@calebkleveter oops, sorry; thanks for fixing `.allTests`!

calebkleveter:
For some reason the test case still isn't running. I have no ides why 😕

MrMage:
> For some reason the test case still isn't running. I have no ides why 😕

Hmm, that's indeed weird.

Unrelated, should we backport this to ConsoleKit 3? (Would probably require opening a `3` branch.)

calebkleveter:
Yeah, I'll take a look at that 👍

penny-coin:
Hey @MrMage, you just merged a pull request, have a coin! 

You now have 14 coins.

calebkleveter:
I created a [`3`](https://github.com/vapor/console/tree/3) branch that we can merge too and tag 3.x.x releases from.

PR #109, opened 2019-05-22
Strict Signature

calebkleveter:
Adds a `.strict` flag to the `CommandRunnable` protocol that causes the arguments and options passed in to be verified against the command's signature before the command is run.

```swift
final class StrictCommand: Command {
    // ...
    static let strict = true
   // ...
}
```


penny-coin:
Hey @calebkleveter, you just merged a pull request, have a coin! 

You now have 516 coins.

PR #122, opened 2019-10-17
Implement enum based CommandError

tkrajacic:
As discussed, this changes CommandError from being a struct to an enum.

The naming of the cases is a bit iffy, and I'll happily change those to better names.

### DIFF
--- a/Sources/ConsoleKit/Command/CommandError.swift
+++ b/Sources/ConsoleKit/Command/CommandError.swift
@@ -1,11 +1,29 @@
 /// Errors working with the `Command` module.
-public enum CommandError: Error, CustomStringConvertible {
+public enum CommandError: Error, Equatable, CustomStringConvertible {
     case missingCommand
     case unknownCommand(_ command: String, available: [String])
     case missingRequiredArgument(_ argument: String)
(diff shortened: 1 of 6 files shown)


Sources/ConsoleKit/Command/CommandError.swift
```diff
@@ -1,47 +1,44 @@
...
-
-    /// See `Debuggable`.
-    public let reason: String
-    
-    /// The name of the attempted command
-    public let command: String?
-    
-    /// All available commands
-    public let availableCommands: [String]
+public enum CommandError: Error, CustomStringConvertible {
```
calebkleveter: I think it would make sense to conform this enum to `Equatable`. It might make your error tests more concise, and with this being a public type, we don't know what dependents on this package will want to do. I expect the compiler will be able to synthesize the conformance, so it shouldn't be too hard.
```suggestion
public enum CommandError: Error, Equatable, CustomStringConvertible {
```

tkrajacic:
Too bad we can't overload enum cases (yet), otherwise we could create a lot nicer names. But for now, I think they are descriptive and sufficient.

This is ready for review.

PR #136, opened 2020-03-28
Allow overriding whether ANSI escape sequences are output

gwynne:
- Particularly useful for implementing a command-line option of the form `--color=yes|no|auto` (corresponding to setting the override to `true`, setting the override to `false`, and leaving it unset, respectively).
- When not building in Xcode, an attempt is made to determine whether output is to a terminal that can support escape codes and they are suppressed if the answer is no.
- Fixed some misaligned whitespace in `--help`'s output.
- An EOF on `stdin` (such as by pressing `Ctrl-D`) no longer causes `Console.choose()` and `Console.confirm()` to go into an infinite loop of printing the prompt over and over.

### DIFF
--- a/Sources/ConsoleKit/Command/CommandGroup.swift
+++ b/Sources/ConsoleKit/Command/CommandGroup.swift
@@ -51,8 +51,8 @@ extension CommandGroup {
     }
 
     private func outputGroupHelp(using context: inout CommandContext) {
-        context.console.output("\("Usage: ", style: .info) \(context.input.executable)", newLine: false)
-        context.console.output("\("<command>", style: .warning) ", newLine: false)
+        context.console.output("\("Usage:", style: .info) \(context.input.executable) ", newLine: false)
+        context.console.output("\("<command>", style: .warning)", newLine: false)
         context.console.print()
 
         if !self.help.isEmpty {
@@ -64,7 +64,7 @@ extension CommandGroup {
         if self.commands.count > 0 {
             context.console.print()
             context.console.output("Commands:".consoleText(.success))
-            for (key, command) in self.commands {
+            for (key, command) in self.commands.sorted(by: { $0.key < $1.key }) {
                 context.console.outputHelpListItem(
                     name: key,
                     help: command.help,
--- a/Sources/ConsoleKit/Console.swift
+++ b/Sources/ConsoleKit/Console.swift
@@ -48,19 +48,6 @@ public protocol Console: class {
     /// - parameters:
     ///     - secure: If `true`, the input should not be shown while it is entered.
     ///
-    /// - returns: the string that was read, or `nil` if EOF was encountered
-    func read(isSecure: Bool) -> String?
-
-    /// Returns a `String` of input read from the `Console` until a line feed character was found.
-    ///
-    ///     let input = console.input(isSecure: false)
-    ///     print(input) // String
-    ///
-    /// - note: The line feed character should not be included.
-    ///
-    /// - parameters:
-    ///     - secure: If `true`, the input should not be shown while it is entered.
-    ///
     /// - returns: the string that was read, or an empty string if EOF was encountered
     func input(isSecure: Bool) -> String
 
--- a/Sources/ConsoleKit/Console.swift
+++ b/Sources/ConsoleKit/Console.swift
@@ -47,6 +47,21 @@ public protocol Console: class {
     ///
     /// - parameters:
     ///     - secure: If `true`, the input should not be shown while it is entered.
+    ///
+    /// - returns: the string that was read, or `nil` if EOF was encountered
+    func read(isSecure: Bool) -> String?
+
+    /// Returns a `String` of input read from the `Console` until a line feed character was found.
+    ///
+    ///     let input = console.input(isSecure: false)
+    ///     print(input) // String
+    ///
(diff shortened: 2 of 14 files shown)


gwynne:
Note: No unit test is included because there isn't really a good way to test this without being able to capture the data sent to `stdout`. I could write code to do that, but it'd be macabre at best. I could also modify `Terminal` to be able to redirect its output (and other) streams, but as this is very unlikely to be in any way useful to Vapor users, it's a lot of extra code to maintain to very little purpose.

Sources/ConsoleKit/Terminal/Terminal.swift
```diff
@@ -13,10 +13,13 @@ public final class Terminal: Console {
...
     /// Dynamically exclude ANSI commands when in Xcode since it doesn't support them.
     internal var enableCommands: Bool {
+        if let override = self.stylizedOutputOverride {
+            return override
+        }
         #if Xcode
             return false
         #else
-            return true
+            return isatty(STDOUT_FILENO) > 0 ? true : false // returns true in Xcode!
```
Mordil: Why not
```suggestion
return isatty(STOUT_FILENO) > 0
```
?

gwynne:
Note: The commit to add EOF-on-stdin protection does break API for people who have their own types that conform to `Console`, but I doubt this is a common enough case to cause any significant issues at this time. Technically this is enough to make this PR `semver-major`, but we can get away with `semver-minor` if this doesn't take too long to merge.

Sources/ConsoleKit/Console.swift
```diff
@@ -47,6 +47,21 @@ public protocol Console: class {
...
+    ///     let input = console.input(isSecure: false)
+    ///     print(input) // String
+    ///
+    /// - note: The line feed character should not be included.
+    ///
+    /// - parameters:
+    ///     - secure: If `true`, the input should not be shown while it is entered.
+    ///
+    /// - returns: the string that was read, or an empty string if EOF was encountered
     func input(isSecure: Bool) -> String
```
MrLotU: Should this be deprecated?

gwynne: I actually marked it deprecated at first, but of course that results in the `console-kit` package itself spewing deprecation errors that users can't do anything about, which is a non-starter. Naturally, Swift still has no way to silence deprecation warnings for such a case despite many requests for it. I figure we can just fix it up in Vapor 5 anyhow.

MrLotU: 👍

Sources/ConsoleKit/Console.swift
```diff
@@ -47,6 +47,21 @@ public protocol Console: class {
     ///
     /// - parameters:
     ///     - secure: If `true`, the input should not be shown while it is entered.
+    ///
+    /// - returns: the string that was read, or `nil` if EOF was encountered
+    func read(isSecure: Bool) -> String?
```
Mordil: Couldn't this be added as an extension that in the next major version will be added to the protocol?

Mordil: Ah, dumb question. I see that there's plenty of internals needed to support the method than can't be done in a protocol extension

Sources/ConsoleKit/Terminal/Terminal.swift
```diff
@@ -13,10 +13,13 @@ public final class Terminal: Console {
 
     /// Dynamically exclude ANSI commands when in Xcode since it doesn't support them.
     internal var enableCommands: Bool {
+        if let override = self.stylizedOutputOverride {
```
MrLotU: should we rename this, since `override` is a swift keyword?

gwynne: Fixed 🙂

Sources/ConsoleKit/Input/Console+Ask.swift
```diff
@@ -14,6 +14,10 @@ extension Console {
     ///     - prompt: Text to display before asking for input.
     ///     - isSecure: See `Console.input(isSecure:)`
     /// - returns: Input `String`. entered in response to the prompt.
+    ///
+    /// - note: If EOF appears from the terminal's input, the result is an empty
```
tanner0101: No longer true.

gwynne: Fixed

tanner0101:
These changes are now available in [4.1.0](https://github.com/vapor/console-kit/releases/tag/4.1.0)

Sources/ConsoleKit/Command/CommandGroup.swift
```diff
@@ -64,7 +64,7 @@ extension CommandGroup {
         if self.commands.count > 0 {
             context.console.print()
             context.console.output("Commands:".consoleText(.success))
-            for (key, command) in self.commands {
+            for (key, command) in self.commands.sorted(by: { $0.key < $1.key }) {
```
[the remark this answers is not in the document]
MrLotU: Does it?  This only sorts commands, #119 seems to be about help options? (might be me missing something though)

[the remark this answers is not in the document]
MrLotU: Nevermind, figured it out. I will close #119

PR #138, opened 2020-05-11
Way awesome new CI

gwynne:


.github/workflows/test.yml
```diff
@@ -2,28 +2,50 @@ name: test
...
-  console-kit_xenial:
-    container: 
-      image: vapor/swift:5.2-xenial-ci
+  console-kit_linux:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image: ['swift:5.2-xenial', 'swift:5.2-bionic',
+                'swiftlang/swift:nightly-5.2-xenial', 'swiftlang/swift:nightly-5.2-bionic',
```
tanner0101, line 11: I'd prefer YML style arrays here if possible. Less syntactical bloat.

.github/workflows/test.yml
```diff
@@ -2,28 +2,50 @@ name: test
...
+      - run: dnf install -y zlib-devel
+        if: ${{ matrix.image == 'swiftlang/swift:nightly-master-centos8' }}
+      - run: yum install -y zlib-devel
+        if: ${{ matrix.image == 'swiftlang/swift:nightly-master-amazonlinux2' }}
+      - uses: actions/checkout@v2
+        with:
+          path: console-kit
+      - uses: actions/checkout@v2
+        with:
+          repository: vapor/vapor
```
tanner0101, line 27: I think it could make sense to have this as a separate test still. We don't really need to compile and test Vapor on every platform. Just one is good enough to make sure we didn't break API or introduce errors. Anything potentially different between operating systems should already be tested within ConsoleKit's unit tests.

.github/workflows/test.yml
gwynne: GitHub claims to prefer the flow collection syntax over block collection for matrix variables, though I suspect that's just a matter of their left hand not knowing that their right hand exists, they're worse with YAML parsers in Actions than Apple is at XML parsers in the iOS kernel. (Yeah, I went there!) I could remove the quotes if you feel better about it that way - they're not necessary in this array as far as I recall.

.github/workflows/test.yml
gwynne: Yeah, I was thinking that myself - but at the same time it's running within the same job so it's not using up extra slots (whereas a separate test would, ironically), and it provides a not-unwelcome backup check of Vapor itself in case the CI workflows don't all match up (and thus far I don't think they ever have yet 😅).

.github/workflows/test.yml
tanner0101: I'm using traditional yml arrays here and it's working fine: https://github.com/vapor/fluent-postgres-driver/blob/gm/.github/workflows/test.yml#L13-L27

Pros of this method are:
- Better git diffs
- Room for comments
- No line wrapping

.github/workflows/test.yml
tanner0101: I think having just one job compiling Vapor and running its unit tests instead of all 12+ jobs doing it would mean CI would run faster. Waiting on CI is the bane of my existence and kills my productivity so anything to help mitigate that is greatly appreciated. If I'm wrong and it wouldn't actually be faster then I'm fine with how it is.

.github/workflows/test.yml
gwynne: I didn't say it didn't _work_, just that GH claims to like it better 😛

.github/workflows/test.yml
gwynne: I mean, is it faster to build just `console-kit` than it is to build `vapor`? Sure. Is the difference particularly significant when using this "all at once" setup instead of the previous "do console-kit and then separately do vapor+console-kit" way? From what I've seen, the GitHub runners are powerful enough that the extra runtime measures around 10 seconds per job. CI via GH Actions takes most of its time in setup/teardown overhead, so the Vapor build and tests would have take about 3x as long as they do now for the difference to matter much.

.github/workflows/test.yml
```diff
@@ -2,28 +2,50 @@ name: test
...
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image: ['swift:5.2-xenial', 'swift:5.2-bionic',
+                'swiftlang/swift:nightly-5.2-xenial', 'swiftlang/swift:nightly-5.2-bionic',
+                'swiftlang/swift:nightly-5.3-xenial', 'swiftlang/swift:nightly-5.3-bionic',
+                'swiftlang/swift:nightly-master-xenial', 'swiftlang/swift:nightly-master-bionic',
+                'swiftlang/swift:nightly-master-focal', 'swiftlang/swift:nightly-master-centos8',
+                'swiftlang/swift:nightly-master-amazonlinux2']
```
gwynne, line 15:
```suggestion
        image:
          - swift:5.2-xenial
          - swift:5.2-bionic
          - swiftlang/swift:nightly-5.2-xenial
          - swiftlang/swift:nightly-5.2-bionic
          - swiftlang/swift:nightly-5.3-xenial
          - swiftlang/swift:nightly-5.3-bionic
          - swiftlang/swift:nightly-master-xenial
          - swiftlang/swift:nightly-master-bionic
          - swiftlang/swift:nightly-master-focal
          - swiftlang/swift:nightly-master-centos8
          - swiftlang/swift:nightly-master-amazonlinux2
```

.github/workflows/test.yml
tanner0101: Screen Shot 2020-05-20 at 11 55 00 AM

From that it seems like testing Vapor is adding a whopping 3 minutes to each test case. The total runtime is 5 minutes so it's more than doubling it. 

Do you mean _total_ runtime is only increased by 10 seconds?

gwynne: TBH, off the top of my head I don't actually recall which number I was looking at. Either way, I'm fine with running just one set of Vapor tests, definitely beats going back and forth over it like this 🙂 Will update the PR shortly.

PR #143, opened 2020-06-01
Fix `Argument` Shell Completion

mliberman:
This fixes the shell completion for label-less `Argument`s in both bash and zsh to correctly reflect the `CompletionAction`.

### DIFF
--- a/Sources/ConsoleKit/Command/Completion.swift
+++ b/Sources/ConsoleKit/Command/Completion.swift
@@ -95,8 +95,8 @@ extension AnyCommand {
 
             if [[ "$COMP_CWORD" -ne \(commandDepth) ]]; then
                 case $prev in
-        \( arguments.map { argument in
-            let label = argument.labels?.values.joined(separator: "|") ?? "*"
+        \( arguments.compactMap { argument in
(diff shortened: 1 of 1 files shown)


Sources/ConsoleKit/Command/Completion.swift
```diff
@@ -297,7 +303,7 @@ extension CompletionAction {
     /// Creates a `CompletionAction` that provides a predefined list of possible values.
     public static func values(_ values: [String]) -> CompletionAction {
         return [
-            .bash: "COMPREPLY=( $(compgen -W \"\(values.joined(separator: " "))\" -- $cur) )",
+            .bash: "COMPREPLY+=( $(compgen -W \"\(values.joined(separator: " "))\" -- $cur) )",
```
mliberman, line 306: The completion values for label-less `Argument`s are _appended_ to `COMPREPLY`, which will already contain any `Option` or `Flag` labels.

calebkleveter:
```
/__w/console-kit/console-kit/Sources/ConsoleKit/Terminal/Terminal.swift:6:8: error: missing required module 'CoreFoundation'
import Foundation
```

This looks fun! I imagine this is just a CI issue and isn't related to your changes though.

PR #161, opened 2021-06-04
Change run() to run() async

finestructure:
Clumsy start at an attempt to fix https://github.com/vapor/vapor/issues/2635

According to what I've read, an `f() async` protocol requirement can be implemented by `f()` - meaning existing protocol conformances should continue to work.

The current code does compile but since I used `DispatchGroup`, errors are not passed through.

The alternative would be to use

```swift
import _NIOConcurrency

let promise = ...eventLoop.makePromise(of: Void.self)
promise.completeWithAsync {
    try await self.run(command, with: CommandContext(console: self, input: input))
}
try promise.futureResult.wait()
```

However, swift-nio isn't currently a package dependency. Maybe it's possible to extract just `promise.completeWithAsync` from it but it wasn't immediately obvious how that could be done.

This might be a complete dead end or perhaps it's a useful start :)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,6 +15,12 @@ let package = Package(
     targets: [
         .target(name: "ConsoleKit", dependencies: [
             .product(name: "Logging", package: "swift-log"),
+        ],
+        swiftSettings: [
+            .unsafeFlags([
+                "-Xfrontend", "-enable-experimental-concurrency",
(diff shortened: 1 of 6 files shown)


finestructure:
Closing this in favour of #164 which seems to be much further along in supporting `func run(...) async throws`.

PR #175, opened 2022-06-23
Modify the`AsyncCommand` implementation `Command` to support asynchronous methods through `EventLoopPromise`, so as to support the `Command` of vapor to support asynchronous methods

josercc:
```swift
@available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
final class TestCommand: AsyncCommand {
    struct Signature: CommandSignature {
        @Argument(name: "foo", help: """
        A foo is required
        An error will occur if none exists
        """)
        var foo: String

        @Option(name: "bar", short: "b", help: """
        Add a bar if you so desire
        Try passing it
        """)
        var bar: String?

        @Flag(name: "baz", short: "B", help: """
        Add a baz if you so desire
        It's just a flag
        """)
        var baz: Bool

        init() { }
    }

    let help: String = "This is a test command"

    func run(using context: CommandContext, signature: Signature) async throws {
        XCTAssertEqual(signature.$foo.name, "foo")
        XCTAssertEqual(signature.$bar.name, "bar")
        XCTAssertEqual(signature.$baz.name, "baz")
        context.console.output("Foo: \(signature.foo) Bar: \(signature.bar ?? "nil")".consoleText(.info))
    }
}
```

Because the bottom layer of `AsyncCommand` implements `Command`, I deleted some unnecessary classes


oscarvgg (approved):
This should have been the implementation of AsyncCommand from the start

t-ae (changes requested):
There's use of `AsyncCommand` left and it blocks compile.
https://github.com/josercc/console-kit/blob/async-command/Sources/ConsoleKitAsyncExample/entry.swift#L11C1-L11C1

Documentation also needs change.
https://github.com/josercc/console-kit/blob/async-command/Sources/ConsoleKit/Command/Async/AsyncCommand.swift#L53

Sources/ConsoleKit/Command/Async/AsyncCommand.swift
```diff
@@ -79,33 +79,21 @@
...
 ///
 #if swift(>=5.5) && canImport(_Concurrency)
+import NIO
+
 @available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
-public protocol AsyncCommand: AnyAsyncCommand {
-    associatedtype Signature: CommandSignature
-    func run(using context: CommandContext, signature: Signature) async throws
+public protocol AsyncCommand: Command {
+    func run(usingAsync context: CommandContext, signature: Signature) async throws
```
t-ae, line 86: This can be `func run(using context: CommandContext, signature: Signature) async throws`.

t-ae (changes requested):

gwynne:
@josercc Can you resolve the conflicts with `main` (preferably by rebase, but whatever method you feel best able to work with is fine)

josercc:
> @josercc Can you resolve the conflicts with `main` (preferably by rebase, but whatever method you feel best able to work with is fine)?

Ok, I'll try it.

josercc:
@gwynne 
Sorry, I have been using Vapor in depth when submitting, but I have stopped using it recently. During the process of merging the code, I found something that needs to be changed. I am afraid of affecting the stability of the upstream Vapor, so I withdraw this commit. At present, I have forgotten the core logic in it.

PR #177, opened 2023-01-20
Add MetadataProvider support

MahdiBM:
This PR:
* Adds `MetadataProvider` support to `ConsoleLogger`.
* Makes it so the the package requires `swift-log` `1.5.1`+.
* Fixes these warnings:
```
[ WARNING ] Attempted to set metadataProvider on ConsoleLogger that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -11,7 +11,7 @@ let package = Package(
         .library(name: "ConsoleKit", targets: ["ConsoleKit"]),
     ],
     dependencies: [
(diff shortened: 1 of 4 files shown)


ktoso (approved):
LGTM! But the lib logging warnings without this change was not intended so I'll see that fixed in swift log, thanks for reporting!

0xTim (approved):
Thanks!

Issue #140, opened 2020-05-20
Logging improvements.


ileitch:
Hey all 👋

I'm new Vapor and loving it so far! My prior development experience is most recently iOS, but before that I was an SRE using Rails. I've discovered that logging in Vapor doesn't quite match my expectations based on my past experiences.

Example 1: Server

```
[ INFO ] GET /favicon.ico
[ ERROR ] Not Found
```

At the `info` log level this output is OK (if a little terse), but at the `warning` level and above it becomes useless since the error line provides no context.

Example 2: Queues

```
[ INFO ] Starting jobs worker (queue: tmdb)
[ INFO ] Dequeing Job:
[ INFO ] Dispatched queue job
[ INFO ] Dequeing Job:
[ INFO ] Dequeing Job:
[ INFO ] Dispatched queue job
[ INFO ] Dequeing Job:
```

This output at least tells me _something_ happened, but otherwise it's mostly useless. I'll discuss log levels more below, but I think it's important to mention that this is the default output for Vapor in the development environment, and therefore what newcomers experience.

Vapor can of course log more information using the metadata mechanism, but at at least in `ConsoleLogger` that's restricted to the `debug` level. This behavior is contrary to my past experiences. For iOS we would log quite liberally at the `info` level, and include those logs in crash reports sent to Crashlytics. Things we'd log at this level included network activity, state machine transitions, events, etc. We logged everything necessary to help us track the users progress through the app to help in diagnosing the crash. In Rails we also used logging liberally.

We could try to define exactly the amount of information that each log level should include, but I think many people have their own preference, so while sensible defaults are great, it'd be great if we could accommodate everyone's personal preference.

Therefore I'd like to put forward some ideas. Please note that the examples provided below aren't serious API suggestions, they're just to convey the concepts.

### 1. Associate each metadata item with a log level

```swift
logger.info("Hello", metadata: [
    .info("job", name),
    .debug("payload", payload)
])
```

Rather than all metadata only being logged at the debug level, this gives us the ability to more finely control individual log line verbosity.

### 2. Define default metadata items, and allow their log level to be customized

Each subsystem (e.g queues, server, etc.) defines a default set of metadata items that are included on every log line output by that subsystem, even for custom lines added by users. For example the queues subsystem might contain:

```swift
struct MetadataItem {
    static var timestamp: Self {
        .init(name: "timestamp", value: { ... })
    }
    static var jobId: Self {
        .init(name: "job-id", value: { ... })
    }
[...]

PR #182, opened 2023-08-19
Implement LoggerFragment system for customizing logger output

semicoleon:
This PR refactors `ConsoleLogger` to break out the functionality into smaller pieces. The goal of this is to allow end users to customize the log output without having to deal with writing a logger themselves. Things like adding a timestamp at the beginning of the log message can be accomplished without having to re-write any of the logic inside `ConsoleKit` by writing a custom `LoggerFragment`. Additionally, since the actual code for constructing the log message is so short, copying it to make modifications inside the message isn't as much of a commitment as copying the whole `ConsoleLogger` implementation is.

This change would at least partially address #140 by making it easier to customize when metadata, etc is logged.

This change also incorporated moving from the deprecated `Logging.LogHandler` protocol method to the newer one which includes the source parameter

### DIFF
--- a/Package@swift-5.8.swift
+++ b/Package@swift-5.8.swift
@@ -1,0 +1,40 @@
+// swift-tools-version:5.8
+import PackageDescription
+
+let package = Package(
+	name: "console-kit",
+	platforms: [
+		.macOS(.v10_15),
+		.iOS(.v13),
+		.watchOS(.v6),
+		.tvOS(.v13),
+	],
+	products: [
+		.library(name: "ConsoleKit", targets: ["ConsoleKit"]),
+	],
+	dependencies: [
+		.package(url: "https://github.com/apple/swift-log.git", from: "1.5.3"),
+		.package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0")
+	],
+	targets: [
+		.target(name: "ConsoleKit", dependencies: [
+			.product(name: "Logging", package: "swift-log"),
+			.product(name: "Atomics", package: "swift-atomics")
+		], swiftSettings: [.enableExperimentalFeature("StrictConcurrency=complete")]),
+		.testTarget(name: "ConsoleKitTests", dependencies: [
+			.target(name: "ConsoleKit"),
+		], swiftSettings: [.enableExperimentalFeature("StrictConcurrency=complete")]),
+		.testTarget(name: "AsyncConsoleKitTests", dependencies: [
+			.target(name: "ConsoleKit"),
+		], swiftSettings: [.enableExperimentalFeature("StrictConcurrency=complete")]),
+		.testTarget(name: "ConsoleKitPerformanceTests", dependencies: [
+			.target(name: "ConsoleKit")
+		], swiftSettings: [.enableExperimentalFeature("StrictConcurrency=complete")]),
+		.executableTarget(name: "ConsoleKitExample", dependencies: [
+			.target(name: "ConsoleKit"),
+		], swiftSettings: [.enableExperimentalFeature("StrictConcurrency=complete")]),
+		.executableTarget(name: "ConsoleKitAsyncExample", dependencies: [
+			.target(name: "ConsoleKit")
+		], swiftSettings: [.enableExperimentalFeature("StrictConcurrency=complete")])
+	]
+)
--- a/Sources/ConsoleKit/Activity/ActivityBar.swift
+++ b/Sources/ConsoleKit/Activity/ActivityBar.swift
@@ -1,5 +1,3 @@
-import Atomics
-
 /// An `ActivityIndicatorType` that renders an activity bar on a single line.
 ///
 ///     Title [=======              ]
@@ -28,7 +26,7 @@ extension ActivityBar {
         let bar: ConsoleText
         switch state {
         case .ready: bar = "[]"
-        case .active(let tick): bar = renderActiveBar(tick: tick, width: Self.width)
+        case .active(let tick): bar = renderActiveBar(tick: tick, width: console.activityBarWidth)
         case .success: bar = "[Done]".consoleText(.success)
         case .failure: bar = "[Failed]".consoleText(.error)
         }
@@ -36,13 +34,29 @@ extension ActivityBar {
     }
 }
 
-/// Defines the width of all `ActivityBar`s in characters.
-private let _width: ManagedAtomic<Int> = ManagedAtomic(25)
-
 extension ActivityBar {
-    /// Defines the width of all `ActivityBar`s in characters.
+    @available(*, deprecated, message: "This value has no effect. Use `console.activityBarWidth` instead.")
     public static var width: Int {
-		get { return _width.load(ordering: .relaxed) }
-		set { _width.store(newValue, ordering: .relaxed)}
+        get { 25 } // deliberately hardcoded value
+        set { } // deliberately ignore new value
+    }
+}
+
+/// Key type for storing the activity bar width in the `userInfo` of the related `Console` without colliding with end user keys.
+struct ActivityBarWidthKey: Hashable, Equatable {
+    func hash(into hasher: inout Hasher) {
+        hasher.combine("ConsoleKit.ActivityBarWidthKey")
+    }
+}
+
+extension Console {
+    public var activityBarWidth: Int {
+        get {
+            self.userInfo[AnyHashable(ActivityBarWidthKey())] as? Int ?? 25
+        }
+        
+        set {
+            self.userInfo[AnyHashable(ActivityBarWidthKey())] = newValue
+        }
     }
 }
--- a/Sources/ConsoleKit/Activity/ActivityBar.swift
+++ b/Sources/ConsoleKit/Activity/ActivityBar.swift
@@ -22,7 +22,7 @@ public protocol ActivityBar: ActivityIndicatorType {
 
 extension ActivityBar {
     /// See `ActivityIndicatorType`.
-    public func outputActivityIndicator(to console: Console, state: ActivityIndicatorState) {
+    public func outputActivityIndicator(to console: any Console, state: ActivityIndicatorState) {
         let bar: ConsoleText
         switch state {
         case .ready: bar = "[]"
@@ -52,11 +52,11 @@ struct ActivityBarWidthKey: Hashable, Equatable {
 extension Console {
     public var activityBarWidth: Int {
         get {
-            self.userInfo[AnyHashable(ActivityBarWidthKey())] as? Int ?? 25
+            self.userInfo[AnySendableHashable(ActivityBarWidthKey())] as? Int ?? 25
         }
         
         set {
-            self.userInfo[AnyHashable(ActivityBarWidthKey())] = newValue
+            self.userInfo[AnySendableHashable(ActivityBarWidthKey())] = newValue
         }
     }
 }
--- a/Sources/ConsoleKit/Activity/ActivityBar.swift
+++ b/Sources/ConsoleKit/Activity/ActivityBar.swift
@@ -1,3 +1,5 @@
+import Atomics
+
 /// An `ActivityIndicatorType` that renders an activity bar on a single line.
 ///
 ///     Title [=======              ]
@@ -35,12 +37,12 @@ extension ActivityBar {
 }
 
 /// Defines the width of all `ActivityBar`s in characters.
-private var _width: Int = 25
+private let _width: ManagedAtomic<Int> = ManagedAtomic(25)
 
 extension ActivityBar {
     /// Defines the width of all `ActivityBar`s in characters.
     public static var width: Int {
-        get { return _width }
-        set { _width = newValue}
+		get { return _width.load(ordering: .relaxed) }
+		set { _width.store(newValue, ordering: .relaxed)}
     }
 }
--- a/Sources/ConsoleKit/Terminal/Terminal.swift
+++ b/Sources/ConsoleKit/Terminal/Terminal.swift
@@ -122,9 +122,13 @@ public final class Terminal: Console, Sendable {
 
     /// See `Console`
     public func report(error: String, newLine: Bool) {
-        let output = newLine ? error + "\n" : error
-        let data = output.data(using: .utf8) ?? Data()
-        FileHandle.standardError.write(data)
+        for c in (newLine ? "\(error)\n" : error).utf8 {
+#if os(Windows)
+            _putc_nolock(CInt(c), stderr)
+#else
+            putc_unlocked(CInt(c), stderr)
+#endif
+        }
     }
 
     /// See `Console`
--- a/Sources/ConsoleKit/Terminal/Terminal.swift
+++ b/Sources/ConsoleKit/Terminal/Terminal.swift
@@ -91,23 +91,23 @@ public final class Terminal: Console, Sendable {
 
     /// See `Console`
     public func output(_ text: ConsoleText, newLine: Bool) {
-		if self.enableCommands {
-			var lines = 0
-			for fragment in text.fragments {
-				let strings = fragment.string.split(separator: "\n", omittingEmptySubsequences: false)
-				for string in strings {
-					let count = string.count
-					if count > size.width && count > 0 && size.width > 0 {
-						lines += (count / size.width) + 1
-					}
-				}
-				/// add line for each fragment
-				lines += strings.count - 1
-			}
-			if newLine { lines += 1 }
-			
-			didOutputLines(count: lines)
-		}
+        if self.enableCommands {
+            var lines = 0
+            for fragment in text.fragments {
+                let strings = fragment.string.split(separator: "\n", omittingEmptySubsequences: false)
+                for string in strings {
+                    let count = string.count
+                    if count > size.width && count > 0 && size.width > 0 {
+                        lines += (count / size.width) + 1
+                    }
+                }
+                /// add line for each fragment
+                lines += strings.count - 1
+            }
+            if newLine { lines += 1 }
+            
+            didOutputLines(count: lines)
+        }
 
         let terminator = newLine ? "\n" : ""
 
--- a/Sources/ConsoleKit/Terminal/Terminal.swift
+++ b/Sources/ConsoleKit/Terminal/Terminal.swift
@@ -9,9 +9,9 @@ import Foundation
 
 /// Generic console that uses a mixture of Swift standard
 /// library and Foundation code to fulfill protocol requirements.
-public final class Terminal: Console {
+public final class Terminal: Console, Sendable {
     /// See `Console`
-    public var userInfo: [AnyHashable: Any]
+    public var userInfo: [AnyHashable: Sendable]
 
     /// Dynamically exclude ANSI commands when in Xcode since it doesn't support them.
     internal var enableCommands: Bool {
--- a/Sources/ConsoleKit/Terminal/Terminal.swift
+++ b/Sources/ConsoleKit/Terminal/Terminal.swift
@@ -91,7 +91,7 @@ public final class Terminal: Console, Sendable {
 
     /// See `Console`
     public func output(_ text: ConsoleText, newLine: Bool) {
-		if !self.enableCommands {
+		if self.enableCommands {
 			var lines = 0
 			for fragment in text.fragments {
 				let strings = fragment.string.split(separator: "\n", omittingEmptySubsequences: false)
--- a/Sources/ConsoleKit/Terminal/Terminal.swift
+++ b/Sources/ConsoleKit/Terminal/Terminal.swift
@@ -6,12 +6,37 @@ import Glibc
 import CRT
 #endif
 import Foundation
+import NIOConcurrencyHelpers
+
+/// A `Sendable` version of the standard library's `AnyHashable` type.
+public struct AnySendableHashable: @unchecked Sendable, Hashable, ExpressibleByStringLiteral {
+    // Note: @unchecked Sendable since there's no way to express that `wrappedValue` is Sendable, even though we ensure that it is in the init.
+    let wrappedValue: AnyHashable
+    
+    public init(_ wrappedValue: any Hashable & Sendable) {
+        self.wrappedValue = AnyHashable(wrappedValue)
+    }
+    
+    public init(stringLiteral value: String) {
+        self.init(value)
+    }
+}
 
 /// Generic console that uses a mixture of Swift standard
 /// library and Foundation code to fulfill protocol requirements.
 public final class Terminal: Console, Sendable {
+    let _userInfo: NIOLockedValueBox<[AnySendableHashable: any Sendable]>
+    
     /// See `Console`
-    public var userInfo: [AnyHashable: Sendable]
+    public var userInfo: [AnySendableHashable: any Sendable] {
+        get {
+            self._userInfo.withLockedValue { $0 }
+        }
+        
+        set {
+            self._userInfo.withLockedValue { $0 = newValue }
+        }
+    }
 
     /// Dynamically exclude ANSI commands when in Xcode since it doesn't support them.
     internal var enableCommands: Bool {
@@ -27,7 +52,7 @@ public final class Terminal: Console, Sendable {
 
     /// Create a new Terminal.
     public init() {
-        self.userInfo = [:]
+        self._userInfo = NIOLockedValueBox([:])
     }
 
     /// See `Console`
--- a/Sources/ConsoleKit/Terminal/Terminal.swift
+++ b/Sources/ConsoleKit/Terminal/Terminal.swift
@@ -91,21 +91,23 @@ public final class Terminal: Console {
 
     /// See `Console`
     public func output(_ text: ConsoleText, newLine: Bool) {
-        var lines = 0
-        for fragment in text.fragments {
-            let strings = fragment.string.split(separator: "\n", omittingEmptySubsequences: false)
-            for string in strings {
-                let count = string.count
-                if count > size.width && count > 0 && size.width > 0 {
-                    lines += (count / size.width) + 1
-                }
-            }
-            /// add line for each fragment
-            lines += strings.count - 1
-        }
-        if newLine { lines += 1 }
-
-        didOutputLines(count: lines)
+		if !self.enableCommands {
+			var lines = 0
+			for fragment in text.fragments {
+				let strings = fragment.string.split(separator: "\n", omittingEmptySubsequences: false)
+				for string in strings {
+					let count = string.count
+					if count > size.width && count > 0 && size.width > 0 {
+						lines += (count / size.width) + 1
+					}
+				}
+				/// add line for each fragment
+				lines += strings.count - 1
+			}
+			if newLine { lines += 1 }
+			
+			didOutputLines(count: lines)
+		}
 
         let terminator = newLine ? "\n" : ""
 
@@ -116,7 +118,6 @@ public final class Terminal: Console {
             output = text.description
         }
         Swift.print(output, terminator: terminator)
-        fflush(stdout)
     }
 
     /// See `Console`
--- a/Sources/ConsoleKit/Utilities/ConsoleLogger.swift
+++ b/Sources/ConsoleKit/Utilities/ConsoleLogger.swift
@@ -2,10 +2,15 @@ import Logging
 
 /// A `LoggerFragment` which implements the default logger message format.
 public func defaultLoggerFragment() -> some LoggerFragment {
-    TimestampFragment().and(LabelFragment().separated(" ")).maxLevel(.trace)
+    LabelFragment().maxLevel(.trace)
         .and(LevelFragment().separated(" ").and(MessageFragment().separated(" ")))
         .and(MetadataFragment().separated(" "))
         .and(SourceLocationFragment().separated(" ").maxLevel(.debug))
+}
+
+/// A `LoggerFragment` which implements the default logger message format with a timestamp at the front.
+public func timestampDefaultLoggerFragment() -> some LoggerFragment {
+    TimestampFragment().and(defaultLoggerFragment().separated(" "))
 }
 
 /// Outputs logs to a `Console` via a `LoggerFragment` pipeline.
@@ -22,7 +27,7 @@ public struct ConsoleFragmentLogger<T: LoggerFragment>: LogHandler, Sendable {
     public var logLevel: Logger.Level
     
     /// The conosle that the messages will get logged to.
-    public let console: Console
+    public let console: any Console
     
     /// The `LoggerFragment` this logger outputs through.
     public var fragment: T
@@ -35,7 +40,7 @@ public struct ConsoleFragmentLogger<T: LoggerFragment>: LogHandler, Sendable {
     ///   - console: The console to log the messages to.
     ///   - level: The minimum level of message that the logger will output. This defaults to `.debug`, the lowest level.
     ///   - metadata: Extra metadata to log with the message. This defaults to an empty dictionary.
-    public init(fragment: T, label: String, console: Console, level: Logger.Level = .debug, metadata: Logger.Metadata = [:]) {
+    public init(fragment: T, label: String, console: any Console, level: Logger.Level = .debug, metadata: Logger.Metadata = [:]) {
         self.fragment = fragment
         self.label = label
         self.metadata = metadata
@@ -43,7 +48,7 @@ public struct ConsoleFragmentLogger<T: LoggerFragment>: LogHandler, Sendable {
         self.console = console
     }
     
-    public init(fragment: T, label: String, console: Console, level: Logger.Level = .debug, metadata: Logger.Metadata = [:], metadataProvider: Logger.MetadataProvider?) {
+    public init(fragment: T, label: String, console: any Console, level: Logger.Level = .debug, metadata: Logger.Metadata = [:], metadataProvider: Logger.MetadataProvider?) {
         self.fragment = fragment
         self.label = label
         self.metadata = metadata
@@ -105,7 +110,7 @@ public struct ConsoleLogger: LogHandler, Sendable {
     public var logLevel: Logger.Level
     
     /// The conosle that the messages will get logged to.
-    public let console: Console
+    public let console: any Console
     
     public var fragment: some LoggerFragment = defaultLoggerFragment()
     
@@ -116,14 +121,14 @@ public struct ConsoleLogger: LogHandler, Sendable {
     ///   - console: The console to log the messages to.
     ///   - level: The minimum level of message that the logger will output. This defaults to `.debug`, the lowest level.
     ///   - metadata: Extra metadata to log with the message. This defaults to an empty dictionary.
-    public init(label: String, console: Console, level: Logger.Level = .debug, metadata: Logger.Metadata = [:]) {
+    public init(label: String, console: any Console, level: Logger.Level = .debug, metadata: Logger.Metadata = [:]) {
         self.label = label
         self.metadata = metadata
         self.logLevel = level
         self.console = console
     }
     
-    public init(label: String, console: Console, level: Logger.Level = .debug, metadata: Logger.Metadata = [:], metadataProvider: Logger.MetadataProvider?) {
+    public init(label: String, console: any Console, level: Logger.Level = .debug, metadata: Logger.Metadata = [:], metadataProvider: Logger.MetadataProvider?) {
         self.label = label
         self.metadata = metadata
         self.logLevel = level
@@ -180,8 +185,29 @@ extension LoggingSystem {
     ///   - console: The console the logger will log the messages to.
     ///   - level: The minimum level of message that the logger will output. This defaults to `.info`.
     ///   - metadata: Extra metadata to log with all messages. This defaults to an empty dictionary.
+    public static func bootstrap(
+        console: any Console,
+        level: Logger.Level = .info,
+        metadata: Logger.Metadata = [:]
+    ) {
+        self.bootstrap(console: console, level: level, metadata: metadata, metadataProvider: nil)
+    }
+    
+    /// Bootstraps a `ConsoleLogger` to the `LoggingSystem`, so that logger will be used in `Logger.init(label:)`.
+    ///
+    ///     LoggingSystem.boostrap(console: console)
+    ///
+    /// - Parameters:
+    ///   - console: The console the logger will log the messages to.
+    ///   - level: The minimum level of message that the logger will output. This defaults to `.info`.
+    ///   - metadata: Extra metadata to log with all messages. This defaults to an empty dictionary.
     ///   - metadataProvider: The metadata provider to bootstrap the logging system with.
-    public static func bootstrap(console: Console, level: Logger.Level = .info, metadata: Logger.Metadata = [:], metadataProvider: Logger.MetadataProvider? = nil) {
+    public static func bootstrap(
+        console: any Console,
+        level: Logger.Level = .info,
+        metadata: Logger.Metadata = [:],
+        metadataProvider: Logger.MetadataProvider? = nil
+    ) {
         self.bootstrap({ (label, metadataProvider) in
             return ConsoleLogger(label: label, console: console, level: level, metadata: metadata, metadataProvider: metadataProvider)
         }, metadataProvider: metadataProvider)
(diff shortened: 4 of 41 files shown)


semicoleon:
That failing check should be covered by `LogHandler`'s extensions

gwynne:
@semicoleon Please merge (or preferably rebase) from main 🙏

semicoleon:
@gwynne done

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
```diff
@@ -1,5 +1,114 @@
...
+#if compiler(>=5.3)
+	return path
+#else
+	let separator: Substring = path.contains("Sources") ? "Sources" : "Tests"
+	return path.split(separator: "/")
+		.split(separator: separator)
+		.last?
+		.joined(separator: "/") ?? path
+#endif
+}
```
gwynne, line 19: The minimum Swift version for the package is now 5.6; you don't need this utility method at all 😅 
```suggestion
```

Sources/ConsoleKit/Utilities/LoggerFragment.swift
```diff
@@ -0,0 +1,281 @@
...
+	/// splits a path on the /Sources/ folder, returning everything after
+	///
+	///     "/Users/developer/dev/MyApp/Sources/Run/main.swift"
+	///     // becomes
+	///     "Run/main.swift"
+	///
+	public var conciseSourcePath: String {
+		ConsoleKit.conciseSourcePath(self.file)
+	}
+
```
gwynne, line 42:
```suggestion
```
Don't need this one either 😅

Sources/ConsoleKit/Utilities/LoggerFragment.swift
```diff
@@ -0,0 +1,281 @@
...
+		output.needsSeparator = true
+	}
+}
+
+/// Writes the file location of the logged message, including the line. This fragment requests a separator for the next fragment.
+public struct FileFragment: LoggerFragment {
+	public init() { }
+	
+	public func write(_ record: inout LogRecord, to output: inout FragmentOutput) {
+		let file = record.conciseSourcePath + ":" + record.line.description
```
gwynne, line 267:
```suggestion
		let file = record.file + ":" + record.line.description
```
(as per above)

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
```diff
@@ -46,66 +157,35 @@ public struct ConsoleLogger: LogHandler {
         set { self.metadata[key] = newValue }
     }
     
-    /// See `LogHandler.log(level:message:metadata:file:function:line:)`.
+    /// See `LogHandler.log(level:message:metadata:source:file:function:line:)`.
```
gwynne, line 160: To avoid a public API break, the original version needs to be provided and call through to the new one.
```suggestion
    /// See `LogHandler.log(level:message:metadata:file:function:line:)`.
    public func log(
        level: Logger.Level,
        message: Logger.Message,
        metadata: Logger.Metadata?,
        file: String,
        function: String,
        line: UInt
    ) {
        self.log(level: level, message: message, metadata: metadata, source: file.prefix(while: { $0 != "/"  }), file: file, function: function, line: line)
    }
    
    /// See `LogHandler.log(level:message:metadata:source:file:function:line:)`.
```

gwynne (changes requested):
Along with the individual nits I've noted, I have a larger concern about the relative performance of this code - logging is often a high-volume output stream, and is always blocking. ConsoleKit's logger is already iffy enough performance-wise; I'm chary of making it worse. Can you do some performance comparisons (ideally in the form of XCTests, but if a microbenchmark is easier, go for it)?

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
semicoleon: The `LogHandler` protocol provides this already, so I don't think this is actually an API break? Unless there's some obscure way to reference the method that can't see the protocol extension method.

https://github.com/apple/swift-log/blob/main/Sources/Logging/LogHandler.swift#L202

semicoleon:
I did some extremely basic comparisons with XCTest's measure method that didn't show any consistent difference, but the tests aren't set up to work in a release build which makes measuring actual performance that way tricky. 

Do you have any examples of how you would want performance testing done?

gwynne:
> Do you have any examples of how you would want performance testing done?

https://github.com/vapor/routing-kit/blob/main/Tests/RoutingKitTests/RouterPerformanceTests.swift gives some medium-okay examples of the basic idea. The critical point is that you _can_ build tests in release mode, but only if you don't use `@testable import` - I've more than once ended up just commenting out entire files worth of unit tests in order to get performance tests to compile (since no one including me ever seems to think to put them in a separate test target or such). In Xcode you can run tests in release by editing the scheme (Cmd-Shift-comma) and changing the configuration for the Test action to Release; from the command line it's `swift test -c release`

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
gwynne: That'd normally be sufficient, but the API breakage checker (see https://github.com/vapor/console-kit/actions/runs/5909305623/job/16030244244?pr=182) is too stupid to realize that, and it's easier to just provide the redundant override than it is to make the checker shut up. That being said, I'm in agreement that it's weird and annoying - I'd be fine with putting `#if CI` around the additional method to satisfy the checker while calling through to the overload in swift-log in other environments.

semicoleon:
I did several runs of that test on the main branch and on the logger-fragment branch, and in release builds I'm seeing the logger-fragment branch is almost twice as fast as main. I'm not really sure what would be responsible for that big of a difference.

1.066 seconds was the initial average I saw with that test on the main branch, but running from terminal seems to be consistently showing an average a little closer to 0.9 seconds. The logger-fragment branch is consistently showing averages on the order of 0.55 seconds.

Sources/ConsoleKit/Utilities/LoggerFragment.swift
```diff
@@ -0,0 +1,271 @@
...
+		
+		guard !allMetadata.isEmpty else { return }
+		
+		output += allMetadata.sortedDescriptionWithoutQuotes.consoleText()
+		output.needsSeparator = true
+	}
+}
+
+/// Writes the file location of the logged message, including the line. This fragment requests a separator for the next fragment.
+public struct FileFragment: LoggerFragment {
```
gwynne, line 253: This should be named `SourceLocationFragment`, or at least `FileAndLineFragment`.

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
```diff
@@ -46,66 +152,35 @@ public struct ConsoleLogger: LogHandler {
...
-    ///     // becomes
-    ///     "Run/main.swift"
-    ///
-    private func conciseSourcePath(_ path: String) -> String {
-        let separator: Substring = path.contains("Sources") ? "Sources" : "Tests"
-        return path.split(separator: "/")
-            .split(separator: separator)
-            .last?
-            .joined(separator: "/") ?? path
-    }
```
gwynne, line 99: I'm willing to bet the lack of this is what's making this branch faster. I'm realizing as I look at this logic that it's not exactly a trivial code path. `split()` in particular is both allocating and `O(n log n)` at _best_ (and can get as bad as `O(2^n)` in a pinch) - the fact that `#fileID` means we don't have to do it at all is probably a large win.

Sources/ConsoleKit/Utilities/LoggerFragment.swift
```diff
@@ -0,0 +1,271 @@
...
+		if let all = self._allMetadata {
+			return all
+		} else {
+			let meta = (self.metadata ?? [:])
+				.merging(self.loggerMetadata, uniquingKeysWith: { (a, _) in a })
+				.merging(self.metadataProvider?.get() ?? [:], uniquingKeysWith: { (a, _) in a })
+			self._allMetadata = meta
+			return meta
+		}
+	}
```
gwynne, line 44: Normally I'd very much applaud the caching here, but in practice does this ever get evaluated more than once per record? (And if it _does_, then the output of the `metadataProvider` could have changed, making the cache invalid - as unlikely as that probably is.)

Tests/ConsoleKitPerformanceTests/ConsoleLoggerPerformanceTests.swift
```diff
@@ -0,0 +1,57 @@
...
+final class TestConsole: Console {
+	var lastOutput: String? = nil
+	var userInfo = [AnyHashable: Any]()
+	
+	func input(isSecure: Bool) -> String {
+		""
+	}
+
+	func output(_ text: ConsoleText, newLine: Bool) {
+		self.lastOutput = text.description + (newLine ? "\n" : "")
```
gwynne, line 21: Try the test without this line - without the additional allocations it implies the numbers might be even more interesting.

Tests/ConsoleKitPerformanceTests/ConsoleLoggerPerformanceTests.swift
```diff
@@ -0,0 +1,57 @@
...
+		self.measure {
+			let console = TestConsole()
+			var logger1 = Logger(label: "codes.vapor.console.1") { label in
+				ConsoleLogger(label: label, console: console)
+			}
+			
+			for _ in 0..<100_000 {
+				logger1.logLevel = .trace
+				logger1[metadataKey: "value"] = "one"
+				logger1.info("Info")
```
gwynne, line 48: Try stressing this further by 1) providing a `metadataProvider` that returns at least one key, and 2) also adding at least one more (unique) metadata item in the actual `.info()` call.

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
gwynne, line 1: While we're here, let's modify `Terminal` to do the following:

1. In `.output(_:newline:)`, skip everything from [lines 94-109](https://github.com/vapor/console-kit/blob/9a12000f4064a2bdc49068d7258292ec1bdc88fc/Sources/ConsoleKit/Terminal/Terminal.swift#L94-L109) when `enableCommands` is `false`. This will save a lot of useless counting of newlines and speed up Vapor's general operation even more.
2. Remove the `fflush(stdout)` on line 119; in modern Swift, `print()` calls `write(2)` directly on the stdout file descriptor, which does not go through the traditional stdio buffer; the flush is pure wasted cycles nowadays.
3. Replace the body of `report(error:newLine:)` with `Swift.print(error, terminator: newLine ? "\n" : "", to: Swift._Stderr)` - this will be a _LOT_ faster and cheaper than routing it through `Data()` and `FileHandle` (though hopefully errors aren't being logged often enough for it to matter).

(Note: Yes, `Swift._Stderr` is technically a misuse of a stdlib SPI - but, like the Mirror workarounds FluentKit uses, it's an SPI that's not going away until at least Swift 6 at worst, and using it here saves an _awful_ lot of unnecessary extra work.)

Sources/ConsoleKit/Utilities/LoggerFragment.swift
gwynne, line 1: Let's also provide a `TimestampFragment`, based on this implementation for generating a timestamp (which I adapted from the one used by `swift-log` for use in production elsewhere), even if we don't use it by default (maybe enable it only at trace level?):

```swift
    func timestamp() -> String {
        String(unsafeUninitializedCapacity: 255) {
            #if canImport(CRT)
                var timestamp = __time64_t()
                var localTime = tm()
                _ = _time64(&timestamp)
                _ = _localtime64_s(&localTime, &timestamp)
            #else
                var timestamp = time(nil)
                var localTime = tm()
                localtime_r(&timestamp, &localTime)
            #endif
            return $0.withMemoryRebound(to: Int8.self) { strftime($0.baseAddress!, $0.count, "%Y-%m-%dT%H:%M:%S%z", &localTime) }
        }
    }
```

gwynne (changes requested):
Several points of issue, but this is nonetheless starting to shape up rather well!

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
semicoleon: Ahh yep, I tried running the test with just the path and that gets averages of about 0.45 seconds. So main without the file path manipulation is a hair faster than the logger-fragment version, but the difference is relatively small. About a microsecond per logged message on my machine, I think?

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
```diff
@@ -1,5 +1,96 @@
...
+/// A `LoggerFragment` which implements the default logger message format.
+public func defaultLoggerFragment() -> some LoggerFragment {
+	LabelFragment().maxLevel(.trace)
+		.and(LevelFragment().separated(" ").and(MessageFragment().separated(" ")))
+		.and(MetadataFragment().separated(" "))
+		.and(FileFragment().separated(" ").maxLevel(.debug))
+}
+
+/// Outputs logs to a `Console` via a `LoggerFragment` pipeline.
+public struct ConsoleFragmentLogger<T: LoggerFragment>: LogHandler {
```
0xTim, line 12: We need to ensure this is Sendable since LogHandlers are now Sendable

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
```diff
@@ -1,5 +1,96 @@
...
+		)
+		
+		self.fragment.write(&record, to: &output)
+
+		self.console.output(output.text)
+	}
+}
+
 /// Outputs logs to a `Console`.
 public struct ConsoleLogger: LogHandler {
```
0xTim, line 95: This should be Sendable as well

0xTim (approved):
Since I'm deep in this world, we need to ensure that new things, especially LogHandlers (because they're marked as such in SwiftLog) are Sendable

0xTim (changes requested):
(Clicked the wrong button!)

Sources/ConsoleKit/Utilities/LoggerFragment.swift
semicoleon: I was thinking it might be common-ish for end users to want to use metadata to filter out messages first thing in the pipeline. In that case the metadata would have to be constructed twice for every message that was actually logged, once to determine whether to filter, and once to actually output it.

Maybe that's not a use case worth optimizing for though.

gwynne: I'd leave it out for now, but make sure you do so in a way that allows putting it back later if needed _without_ breaking public API.

Tests/ConsoleKitPerformanceTests/ConsoleLoggerPerformanceTests.swift
semicoleon: main: 0.356 seconds
logger-fragment: 0.483 seconds

Tests/ConsoleKitPerformanceTests/ConsoleLoggerPerformanceTests.swift
semicoleon: main: 0.491
logger-fragment: 0.637

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
semicoleon: 3 here is giving me a build error with Swift 5.8.1, I'm not sure what's happening there or whether I can fix it

Sources/ConsoleKit/Utilities/LoggerFragment.swift
semicoleon: `String(unsafeUninitializedCapacity:)` requires macOS 11.0, should we fall back to initialized capacity or are we comfortable bumping the required version?

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
semicoleon: Does this imply `LoggerFragment` should be `Sendable`? It looks like doing that would require making `ConsoleText` `Sendable` as well.

semicoleon: Actually even adding the explicit `Sendable` to just the log handlers produces a warning about `Console` not being `Sendable`

0xTim: Yeah and so starts the rabbit hole thread pulling of Sendable. We either need to wrap stuff in `NIOLoopBound` if we can guarantee that we're going to be on an event loop or just make it Sendable

Sources/ConsoleKit/Utilities/LoggerFragment.swift
0xTim: We'll need to fallback. Bumping to 11 is a major breaking change and unfortunately that breaks Vapor

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
0xTim: Looking at it, ConsoleText (and the types it pulls in) can all be marked as Sendable without any issue. You can slap a Sendable on Console and see what breaks

0xTim: Terminal will be the issue because of the user info dictionary that's `Any`. I'm happy to make that `Sendable` instead as that just produces warnings for types that aren't Sendable

semicoleon: I went ahead and pushed some changes. I'm still getting a warning on `userInfo` since it's mutable. Is adding a dependency on NIO worth it for getting the locked box? Should I use NSLock? 

I also had to modify `VeryUnsafeMutableSigAtomicBufferPointer`. The data structure itself is actually immutable but it was using a subscript to mutate the underlying pointer which meant the global variable had to be `var` which strict concurrency checking didn't like. I just made manual methods for getting and setting. That hurts the ergonomics pretty badly, but it's 1. an implementation detail and 2. not used very much. A closure accepting an inout param would have been better, but I wasn't confident that wouldn't run afoul of the strongly worded warnings on the type so I just settled for the simplest option for now.

Sources/ConsoleKit/Terminal/Terminal.swift
```diff
@@ -91,21 +91,23 @@ public final class Terminal: Console {
...
-                    lines += (count / size.width) + 1
-                }
-            }
-            /// add line for each fragment
-            lines += strings.count - 1
-        }
-        if newLine { lines += 1 }
-
-        didOutputLines(count: lines)
+		if !self.enableCommands {
```
gwynne, line 94: This is backwards; we want to do this only if commands are _enabled_, not disabled 😅. Not your fault; my wording when I suggested this was more than a little confusing, even to myself 😆.
```suggestion
		if self.enableCommands {
```

Sources/ConsoleKit/Activity/ActivityBar.swift
```diff
@@ -35,12 +37,12 @@ extension ActivityBar {
 }
 
 /// Defines the width of all `ActivityBar`s in characters.
-private var _width: Int = 25
+private let _width: ManagedAtomic<Int> = ManagedAtomic(25)
```
gwynne, line 40: An atomic is the wrong primitive here; it doesn't yield the correct semantics. Mind you, the entire way this API works is so appallingly wrong that I'm tempted to say "let's break semver-major rules just this once to fix it" - if we take the value from the `userInfo` of the `Console` given in `outputActivityIndicator()`, we avoid this whole mss; the only caveat to doing that is it's not technically a "global" value anymore and the API contract can't be correctly maintained; we could technically avoid the semver-major break by doing this:
```swift
extension ActivityBar {
    @available(*, deprecated, message: "This value has no effect. Use `console.userInfo[\"activityBarWidth\"]` instead.")
    public static var width: Int {
        get { 25 } // deliberately hardcoded value
        nonmutating set { } // deliberately ignore new value
    }
}
```

Package@swift-5.8.swift
```diff
@@ -0,0 +1,40 @@
...
+		.iOS(.v13),
+		.watchOS(.v6),
+		.tvOS(.v13),
+	],
+	products: [
+		.library(name: "ConsoleKit", targets: ["ConsoleKit"]),
+	],
+	dependencies: [
+		.package(url: "https://github.com/apple/swift-log.git", from: "1.5.3"),
+		.package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0")
```
gwynne, line 17: 1. You didn't add this dependency to `Package.swift`, meaning the package won't build in 5.7, but more importantly,
2. You don't need this package anyway (see my comments in `ActivityBar.swift`)

Sources/ConsoleKit/Activity/ActivityBar.swift
```diff
@@ -35,12 +37,12 @@ extension ActivityBar {
...
 
 extension ActivityBar {
     /// Defines the width of all `ActivityBar`s in characters.
     public static var width: Int {
-        get { return _width }
-        set { _width = newValue}
+		get { return _width.load(ordering: .relaxed) }
+		set { _width.store(newValue, ordering: .relaxed)}
```
gwynne, line 46: In addition to my remarks above, both `LoadingBar.swift` (line 45) and `ProgressBar.swift` (lines 54 and 55) incorrectly refer to the global property instead of the `width` parameter, this should be fixed.

Sources/ConsoleKit/Utilities/LoggerFragment.swift
```diff
@@ -0,0 +1,329 @@
...
+				_ = _localtime64_s(&localTime, &timestamp)
+			#else
+				var timestamp = time(nil)
+				var localTime = tm()
+				localtime_r(&timestamp, &localTime)
+			#endif
+			return $0.withMemoryRebound(to: Int8.self) { strftime($0.baseAddress!, $0.count, "%Y-%m-%dT%H:%M:%S%z", &localTime) }
+		}
+	}
+
```
gwynne, line 297: There's no need for providing the alternative approach; just always use the 10.15-compatible version.
```suggestion
        output += self.timestamp().consoleText()
        output.needsSeparator = true
    }	
```
(Also, please set your text editor to use spaces instead of tabs for indentation 😅)

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
gwynne: What is the error?

gwynne: Nevermind, I see the problem. Do this instead:
```swift
    public func report(error: String, newLine: Bool) {
        for c in (newLine ? "\(error)\n" : error).utf8 {
            #if os(Windows)
            _putc_nolock(CInt(c), stderr)
            #else
            putc_unlocked(CInt(c), stderr)
            #endif
        }
    }
```
(This mimics how `SwiftPrivate._Stderr` works in practice.)

semicoleon:
The API breakage checker appears to still be failing even with the `#if CI` method. I think the workflow needs to be something like `swift package diagnose-api-breaking-changes -Xswiftc -DCI origin/main` for that to work.

But also the `Sendable` changes are being marked as API breakage so it's going to fail either way I think

gwynne:
> But also the `Sendable` changes are being marked as API breakage so it's going to fail either way I think

Yep, you're right on target there - there's no point trying to satisfy it now, just ditch the extra declaration.

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
```diff
@@ -1,7 +1,98 @@
 import Logging
 
+/// A `LoggerFragment` which implements the default logger message format.
+public func defaultLoggerFragment() -> some LoggerFragment {
+	TimestampFragment().and(LabelFragment().separated(" ")).maxLevel(.trace)
```
gwynne: @0xTim Hm. `.debug` or `.trace` for the timestamps? We're already using `.debug` as the minimum for logging source location...

0xTim: Can we make it configurable? Most logging systems will add their own timestamps but it comes up often enough that it would be nice to turn it on easily

semicoleon: If we're leaning towards making it configurable, we may as well just leave the timestamp out from the default one. 

All you have to do to add a timestamp to the default log message now is

```swift
LoggingSystem.bootstrap(
    fragment: TimestampFragment().separated(" ").and(defaultLoggerFragment()),
    console: Terminal()
)
```

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
semicoleon: @0xTim do you have thoughts on what to do for the `userInfo` fields? Using a lock with separate getters and setters sounds messy but I'm not sure if there's a better solution with this API.

0xTim: Ah because it's on a class. Yeah wrap it in `NIOLockedValueBox` in a new property called `_userInfo` and then make `userInfo` a computed property that accessed the underlying value via the lock to maintain the public API

semicoleon: @0xTim ConsoleKit doesn't currently depend on NIO, so we'd have to add it as a dependency to use NIOLockedValueBox.

gwynne: That might be a road we want to go down anyway (adding the NIO dependency), since ideally ConsoleKit's API should support Concurrency (and to a certain extent, pre-Concurrency ELF mechanisms).

gwynne: It would also help ease the planned (eventual) transition to fully deprecating the existing `Command` definitions in favor of first-class support for `swift-argument-parser`.

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
```diff
@@ -116,12 +178,29 @@ extension LoggingSystem {
...
-            return ConsoleLogger(label: label, console: console, level: level, metadata: metadata)
-        }
+    ///   - level: The minimum level of message that the logger will output. This defaults to `.info`.
+    ///   - metadata: Extra metadata to log with all messages. This defaults to an empty dictionary.
+    ///   - metadataProvider: The metadata provider to bootstrap the logging system with.
+    public static func bootstrap(console: Console, level: Logger.Level = .info, metadata: Logger.Metadata = [:], metadataProvider: Logger.MetadataProvider? = nil) {
+        self.bootstrap({ (label, metadataProvider) in
+            return ConsoleLogger(label: label, console: console, level: level, metadata: metadata, metadataProvider: metadataProvider)
+        }, metadataProvider: metadataProvider)
+    }
```
gwynne, line 188: Nope, this time you _did_ break public API, there's no `Logger`-provided extension that covers the non-`metadataProvider` version for this one. You have to provide the original version instead of defaulting:
```suggestion
    ///   - level: The minimum level of message that the logger will output. This defaults to `.info`.
    ///   - metadata: Extra metadata to log with all messages. This defaults to an empty dictionary.
    public static func bootstrap(console: Console, level: Logger.Level = .info, metadata: Logger.Metadata = [:]) {
        self.bootstrap(console: console, level: level, metadata: metadata, metadataProvider: nil)
    }
    
    /// Bootstraps a `ConsoleLogger` to the `LoggingSystem`, so that logger will be used in `Logger.init(label:)`.
    ///
    ///     LoggingSystem.boostrap(console: console)
    ///
    /// - Parameters:
    ///   - console: The console the logger will log the messages to.
    ///   - level: The minimum level of message that the logger will output. This defaults to `.info`.
    ///   - metadata: Extra metadata to log with all messages. This defaults to an empty dictionary.
    ///   - metadataProvider: The metadata provider to bootstrap the logging system with.
    public static func bootstrap(console: Console, level: Logger.Level = .info, metadata: Logger.Metadata = [:], metadataProvider: Logger.MetadataProvider?) {
        self.bootstrap({ (label, metadataProvider) in
            ConsoleLogger(label: label, console: console, level: level, metadata: metadata, metadataProvider: metadataProvider)
        }, metadataProvider: metadataProvider)
    }
```

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
0xTim: Yeah I have no problem adding it as a dependency. I can't imagine there are many use cases where it's not pulled in by something else

0xTim: (That being said, in an ideal world the next major release won't need it and have a full Concurrency support)

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
```diff
@@ -116,12 +178,29 @@ extension LoggingSystem {
...
+    ///   - fragment: The logger fragment which will be used to build the logged messages.
+    ///   - console: The console the logger will log the messages to.
+    ///   - level: The minimum level of message that the logger will output. This defaults to `.info`.
+    ///   - metadata: Extra metadata to log with all messages. This defaults to an empty dictionary.
+    ///   - metadataProvider: The metadata provider to bootstrap the logging system with.
+    public static func bootstrap<T: LoggerFragment>(fragment: T, console: Console, level: Logger.Level = .info, metadata: Logger.Metadata = [:], metadataProvider: Logger.MetadataProvider? = nil) {
+        self.bootstrap({ (label, metadataProvider) in
+            return ConsoleFragmentLogger(fragment: fragment, label: label, console: console, level: level, metadata: metadata, metadataProvider: metadataProvider)
+        }, metadataProvider: metadataProvider)
     }
```
gwynne, line 204: Since we're bumping Vapor to require Swift 5.7 for `Sendable` support, let's do the same here (change the `swift-tools-version` in `Package.swift` from 5.6 to 5.7, the CI will adjust automatically). That confers the bonus of the ability to write this method like so:
```suggestion
    public static func bootstrap(
        fragment: some LoggerFragment,
        console: any Console,
        level: Logger.Level = .info,
        metadata: Logger.Metadata = [:],
        metadataProvider: Logger.MetadataProvider? = nil
    ) {
        self.bootstrap({
            ConsoleFragmentLogger(
                fragment: fragment,
                label: $0,
                console: console,
                level: level,
                metadata: metadata,
                metadataProvider: $1
            )
        }, metadataProvider: metadataProvider)
    }
```
Which, while I'm coming up with more annoying things for you to do, reminds me - add `.enableUpcomingFeature("ExistentialAny")` to the `swiftSettings` for the various targets in `Package@swift-5.9.swift` and fix all the resulting error (which will consist of adding `any` in a whole pile of places). Don't worry, the `any` syntax was introduced in 5.6, it'll build without version conditionals; the feature flag just didn't come along until 5.8.

I would recommend also adding these (which combined with the rest would give you the same set I'm using for the updated MySQLNIO), but I'll let that be optional for now:
```swift
                .enableUpcomingFeature("ForwardTrailingClosures"),
                .enableUpcomingFeature("ConciseMagicFile"),
```

gwynne (changes requested):

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
semicoleon: Is adding an argument with a default value not backwards compatible under some circumstances?

gwynne: Dear Gods, here we go again...! 😂 Fortunately, @MahdiBM gave a quite detailed explanation in his pitch about fixing the problem in the language: https://forums.swift.org/t/pitch-adding-new-parameters-w-default-values-should-not-break-api/66847

Sources/ConsoleKit/Utilities/ConsoleLogger.swift
semicoleon: Ahh `AnyHashable` isn't `Sendable`. Is there a way to work around that other than creating a new `@unchecked Sendable` wrapper?

0xTim: Can we not use `Hashable & Sendable`? Or do we need to type-erase it because old swift and multiple types?

gwynne: Even in Swift 5.7 you can't just write `Dictionary<some Hashable & Sendable, any Sendable>`, and I don't believe `any Hashable` works as a key either (though arguably the underlying category theory says it should). But you can write `any (AnyHashable & Sendable)`.

gwynne: Ah, right, sorry, my mistake 🤦‍♀️

gwynne: I think our only option that retains public API compatibility and avoids Concurrency warnings is to trick the compiler by redeclaring `AnyHashable`. It'll be a blighted mess, but it might work:
```swift
/// Ref: https://github.com/apple/swift/blob/main/stdlib/public/core/AnyHashable.swift
@_spi(AnyHashableSendabilityWorkaround)
public struct AnyHashable: Hashable, CustomStringConvertible, CustomDebugStringConvertible, CustomReflectable, _HasCustomAnyHashableRepresentation, @unchecked Sendable {
    var base: Swift.AnyHashable
    public init<H: Hashable & Sendable>(_ base: H) { self.base = .init(base) }
    public var base: Any { self.base.base }
    public static func == (lhs: ConsoleKit.AnyHashable, rhs: ConsoleKit.AnyHashable) -> Bool { lhs.base == rhs.base }
    public var hashValue: Int { self.base.hashValue }
    public func hash(into hasher: inout Hasher) { self.base.hash(into: &hasher) }
    public func _rawHashValue(seed: Int) -> Int { self.base._rawHashValue(seed: seed) }
    public var description: String { self.base.description }
    public var debugDescription: String { self.base.debugDescription }
    public var customMirror: Mirror { self.base.customMirror }
    @_alwaysEmitIntoClient public __consuming func _toCustomAnyHashable() -> AnyHashable? { self.base._toCustomAnyHashable() }
}

Comment (gwynne) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: Technically this is still breaking since non-`Sendable` inputs will no longer be accepted... we'll probably have to remove that requirement and thus _deliberately_ and completely bypass the entire `Sendable` checking machinery, but I don't see much else in the way of options.

Comment (0xTim) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: It's worth checking this. Will they not be accepted or will they produce a warning? Because I'd assume that's the case

Comment (semicoleon) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: You can't use an @_spi type in a public interface can you? Even if we did that it would still break client code because they wouldn't be the same `AnyHashable` type I think?

Comment (gwynne) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: No, they'll be rejected; it's an explicit protocol conformance constraint, not an implied sendability violation.

Comment (gwynne) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: > You can't use an @_spi type in a public interface can you? Even if we did that it would still break client code because they wouldn't be the same `AnyHashable` type I think?

Damn, you're right 🤦‍♀️ And we can't hide it by giving it a leading underscore either...

Comment (gwynne) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: Actually, we don't have to hide it, come to think of it. The worst we'd ever have to do, assuming the stdlib version gains Sendabiltiy at some point, is change it to `public typealias AnyHashable = Swift.AnyHashable` - since the public API is being mimicked method for method, as long as it can still be referred to by module-qualified name, it doesn't matter which version people get (aside from the sendable question)

Comment (semicoleon) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: Well the names will conflict if ours is public. And if an API was previously passing a Swift.AnyHashable as a key directly that would break since that would no longer be a compatible type.

This entire API is extremely not `Sendable` friendly 😅

Comment (gwynne) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: Tim and I are currently discussing the best approach, bear with us 😅

Comment (0xTim) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: @semicoleon ok I think we've decided on creating an `AnySendableHashable` type that ensures the `Hashable` types are `Sendable`. This should be fine for most keys unless people are doing something silly

**Comment by gwynne** [comment]
@semicoleon On a completely unrelated note, are you on the [Vapor Discord server](https://discord.gg/vapor)?

**Comment by gwynne** [comment]
> I am!

I'm starting to feel like I'll regret this, but what's your name there, if you don't mind me asking? 😅

**Comment by semicoleon** [comment]
I'm skiingcows on Discord!

**Comment by gwynne** [comment]
> I'm skiingcows on Discord!

And you haven't chatted in the Vapor server since 2020, at least according to the server search - no wonder I had to ask! 😆

**Comment by gwynne** [comment]
FWIW, I asked because I'd love to be able to give you credit in front of the community and invite you to the Contributors channel there 🙂. (Yep, that means this PR is _that_ good, even before it's quite finished! 😅)

File: Sources/ConsoleKit/Utilities/ConsoleLogger.swift (hunk 10 above)
Comment (semicoleon) [reply Sources/ConsoleKit/Utilities/ConsoleLogger.swift]: I think I have all of the (non-example code) Sendable warning wrangled now except for this one
```text
console-kit/Sources/ConsoleKit/Activity/ActivityIndicator.swift:80:41: warning: type 'any DispatchSourceTimer' does not conform to the 'Sendable' protocol
        self._timer = NIOLockedValueBox(DispatchSource.makeTimerSource(flags: [], queue: self.queue))
                                        ^
Dispatch.DispatchSourceTimer:1:17: note: protocol 'DispatchSourceTimer' does not conform to the 'Sendable' protocol
public protocol DispatchSourceTimer : DispatchSourceProtocol {
```

I would guess that basically all of the types in libdispatch should probably be Sendable, so I'm not sure if this is an oversight on the part of the library or a larger issue.

gwynne: You should be able get around this by writing it this way:
```swift
        self._timer = NIOLockedValueBox(DispatchSource.makeTimerSource(flags: [], queue: self.queue) as DispatchSource)
```
(you may need `as!` instead of `as`)

The trick is that while the `DispatchSource` class itself is `Sendable`, neither `DispatchSourceProtocol` nor the various subprotocols which represent the various specific source types in Swift (such as `DispatchSourceTimer`) follow suit. I believe this is an oversight in the API.

semicoleon:
I *think* that's all of the comments up to this point addressed

Sources/ConsoleKit/Utilities/LoggerFragment.swift
```diff
@@ -0,0 +1,307 @@
...
+public struct SourceLocationFragment: LoggerFragment {
+    public init() { }
+    
+    public func write(_ record: inout LogRecord, to output: inout FragmentOutput) {
+        let file = record.file + ":" + record.line.description
+        output += "(" + file.consoleText() + ")"
+        output.needsSeparator = true
+    }
+}
+
```
gwynne, line 267: Can we also add a `SourceModuleFragment` for rendering the `source` field of the `LogRecord`? It shouldn't be part of the default, but we should provide it.

Sources/ConsoleKit/Utilities/LoggerFragment.swift
```diff
@@ -0,0 +1,307 @@
...
+#else
+        var timestamp = time(nil)
+        let localTime = localtime(&timestamp)
+        strftime(&buffer, buffer.count, "%Y-%m-%dT%H:%M:%S%z", localTime)
+#endif
+        return buffer.withUnsafeBufferPointer {
+            $0.withMemoryRebound(to: CChar.self) {
+                String(cString: $0.baseAddress!)
+            }
+        }
```
gwynne, line 295: I forgot to suggest before to replace this with this much more efficient (and, on macOS and Linux, thread-safe) version:
```suggestion
        withUnsafeTemporaryAllocation(of: CChar.self, capacity: 255) {
            #if os(Windows)
                var timestamp = __time64_t()
                var localTime = tm()
                _ = _time64(&timestamp)
                _ = _localtime64_s(&localTime, &timestamp)
            #else
                var timestamp = time(nil)
                var localTime = tm()
                localtime_r(&timestamp, &localTime)
            #endif
            guard strftime($0.baseAddress!, $0.count, "%Y-%m-%dT%H:%M:%S%z", &localTime) > 0 else {
                return "<unknown>"
            }
            return String(cString: $0.baseAddress!)
        }
```

gwynne (approved):
Apologies for the long delay in re-reviewing this - only a couple of small things left to clean up, and as neither is critical, I'm approving the PR now. If you want to address the last couple of items, I'd rather merge this with them than without, but I'd rather merge this without them than not at all, and definitely sooner than later! 😅

semicoleon:
I realized I didn't write a test for TimestampFragment, which really probably should have one. I tweaked that type a bit to make it a little easier to test (though apparently strftime doesn't have a good way to override the timestamp part of the `struct tm` which is annoying)

0xTim (approved):
I think we're good on my end once @gwynne 's comments have been addressed

PR #192, opened 2023-11-13
Split ConsoleKit into two functional targets

gwynne:
ConsoleKit has been split into two separate targets:

- `ConsoleKitTerminal` contains the logic for interacting with a console (I/O, color support, other ANSI commands, etc.).
- `ConsoleKitCommands` contains the functionality for handling argument parsing, which is now soft-deprecated in favor of `swift-argument-parser` when possible.

The existing `ConsoleKit` target is now an umbrella import, similar to the function of the `NIO` target in `swift-nio`; existing code should be unaffected.

Several additional minor cleanups were also made in the process of this split.


Sources/ConsoleKitCommands/Async/AsyncCommandGroup.swift
```diff
@@ -60,8 +62,8 @@ extension AsyncCommandGroup {
             context.console.print(self.help)
         }
 
-        let padding = self.commands.map { $0.key }.longestCount + 2
-        if self.commands.count > 0 {
+        let padding = (self.commands.max(by: { $0.key.count < $1.key.count })?.key.count ?? 0) + 2
```
MahdiBM, line 65:
```suggestion
        let padding = (self.commands.map(\.key.count).max() ?? 0) + 2
```

Sources/ConsoleKitCommands/CommandGroup.swift
```diff
@@ -60,8 +62,8 @@ extension CommandGroup {
             context.console.print(self.help)
         }
 
-        let padding = self.commands.map { $0.key }.longestCount + 2
-        if self.commands.count > 0 {
+        let padding = (self.commands.max(by: { $0.key.count < $1.key.count })?.key.count ?? 0) + 2
```
MahdiBM, line 65:
```suggestion
        let padding = (self.commands.map(\.key.count).max() ?? 0) + 2
```

Sources/ConsoleKitCommands/Command.swift
```diff
@@ -138,8 +140,8 @@ extension CommandSignature {
             + self.arguments.map { $0.name }
             + self.flags.map { $0.name }
 
-        let padding = names.longestCount + 2
-        if self.arguments.count > 0 {
+        let padding = (names.max(by: { $0.count < $1.count })?.count ?? 0) + 2
```
MahdiBM, line 143:
```suggestion
        let padding = (names.map(\.count).max() ?? 0) + 2
```

Sources/ConsoleKitCommands/Utilities.swift
```diff
@@ -0,0 +1,29 @@
...
+extension Array {
+    /// Pops the first element from the array.
+    mutating func popFirst() -> Element? {
+        guard let pop = first else {
+            return nil
+        }
+        self = Array(dropFirst())
+        return pop
+    }
+}
```
MahdiBM, line 13:
```suggestion
extension Array {
    /// Pops the first element from the array.
    mutating func popFirst() -> Element? {
        guard let pop = self.first else {
            return nil
        }
        self = Array(self.dropFirst())
        return pop
    }
}
```

MahdiBM (changes requested):

Sources/ConsoleKitCommands/Async/AsyncCommandGroup.swift
MahdiBM: BTW this is a suggestion for performance, not a nit. It just happens to be cleaner too.

Sources/ConsoleKitAsyncExample/entrypoint.swift
```diff
@@ -16,7 +16,6 @@ struct AsyncExample {
             try await console.run(group, input: input)
         } catch let error {
             console.error("\(error)")
-            exit(1)
```
0xTim, line 19: Why not `exit(1)` anymore?

Sources/ConsoleKitCommands/Docs.docc/index.md
```diff
@@ -0,0 +1,13 @@
+# ``ConsoleKit``
```
0xTim, line 1:
```suggestion
# ``ConsoleKitCommands``
```

Sources/ConsoleKitCommands/Docs.docc/index.md
```diff
@@ -0,0 +1,13 @@
...
+
+@Metadata {
+    @TitleHeading(Package)
+}
+
+ConsoleKit provides utilities for interacting with a console via a Swift application. It provides:
+
+* A ``Command`` type for writing commands with arguments and flags
+* Utilities for sending and receiving text to a terminal
+* A [Swift Log](https://github.com/apple/swift-log) implementation for a ``Logger`` that outputs to the console
```
0xTim, line 11: This lives in the other package

Sources/ConsoleKitTerminal/Docs.docc/index.md
```diff
@@ -0,0 +1,13 @@
+# ``ConsoleKit``
```
0xTim, line 1:
```suggestion
# ``ConsoleKitTerminal``
```

Sources/ConsoleKitTerminal/Docs.docc/index.md
```diff
@@ -0,0 +1,13 @@
...
+
+@Metadata {
+    @TitleHeading(Package)
+}
+
+ConsoleKit provides utilities for interacting with a console via a Swift application. It provides:
+
+* A ``Command`` type for writing commands with arguments and flags
```
0xTim, line 9: Other package

0xTim (approved):
Documentation needs updating otherwise mainly LGTM

ptoffy (approved):
Nothing to add to the other reviews

Sources/ConsoleKitAsyncExample/entrypoint.swift
gwynne: 'cause it had no meaningful effect, and exiting with a nonzero status is arguably a questionable behavior to provide as an example (especially when `exit(1)` isn't a fully platform-agnostic or even necessarily correct means of doing so)

PR #193, opened 2023-11-16
Fix `AnySendableHashable` regression

0xTim:
`AnySendableHashable` introduced a breaking change where trying to pass concrete types that conform to `Sendable` and `Hashable` but the compiler couldn't infer they could be passed around as `AnySendableHashable`. This fixes the accesses for any dictionary (including `userInfo`) to fix an API break in https://github.com/vapor/console-kit/releases/tag/4.8.0

### DIFF
--- a/Sources/ConsoleKit/Utilities/AnySendableHashable.swift
+++ b/Sources/ConsoleKit/Utilities/AnySendableHashable.swift
@@ -17,3 +17,10 @@ extension AnySendableHashable: CustomStringConvertible, CustomDebugStringConvert
     public var debugDescription: String { self.wrappedValue.debugDescription }
     public var customMirror: Mirror { self.wrappedValue.customMirror }
 }
+
+extension Dictionary where Key == AnySendableHashable {
+    public subscript(key: some Hashable & Sendable) -> Value? {
+        get { self[AnySendableHashable(key)] }
+        set { self[AnySendableHashable(key)] = newValue }
+    }
+}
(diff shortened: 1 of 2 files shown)


Sources/ConsoleKit/Utilities/AnySendableHashable.swift
```diff
@@ -17,3 +17,10 @@ extension AnySendableHashable: CustomStringConvertible, CustomDebugStringConvert
     public var debugDescription: String { self.wrappedValue.debugDescription }
     public var customMirror: Mirror { self.wrappedValue.customMirror }
 }
+
+extension Dictionary where Key == AnySendableHashable {
+    public subscript(key: some Hashable & Sendable) -> Value? {
+        get { self[AnySendableHashable(key)] }
+        set { self[AnySendableHashable(key)] = newValue }
```
gwynne, line 24:
```suggestion
        @inlinable get { self[AnySendableHashable(key)] }
        @inlinable set { self[AnySendableHashable(key)] = newValue }
```
I'm just barely willing to not declare it `@_transparent` outright 😆

Add `@inlinable` to `AnySendableHashable`'s other initializers as well, and mark the `wrappedValue` property `@usableFromInline`. Should yield at least a minor speedup for apps with large numbers of routes.

gwynne (changes requested):

gwynne (approved):

PR #198, opened 2024-05-08
Support compiling against Musl

simonjbeaumont:
In order to compile this package against Musl libc, we need to not conflate `#if os(Linux)` with the ability to `import Glibc`. The common pattern for this is to use the `#if canImport` directive.

This patch adds support for Musl by replacing the `Glibc` import conditional alongside a new conditional import of `Musl`.

It also updates the `linux_readpassphrase` implementation to use the appropriate `sigaction` handler field for Musl.

### DIFF
--- a/Sources/ConsoleKitTerminal/Terminal/Terminal.swift
+++ b/Sources/ConsoleKitTerminal/Terminal/Terminal.swift
@@ -50,7 +50,7 @@ public final class Terminal: Console, Sendable {
         didOutputLines(count: 1)
         if isSecure {
             var pass = ""
-#if canImport(Darwin) || canImport(Glibc) || os(Android)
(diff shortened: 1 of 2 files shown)


0xTim (approved):
Thanks for this @simonjbeaumont !

Do you know if there are Swift images that link against musl for us to point to in CI to make sure we don't introduce any changes in the future that could break this again?

gwynne:
The usual test environment for Musl-based Linux would be the Arch Linux images. They actively maintain a working Swift toolchain in AUR which IIRC is fairly easy to install (though it wasn't quite one-step last time I tried it).

0xTim:
Going to merge this as the CI story is a bit more complicated than initially thought

PR #205, opened 2024-10-20
Fix Windows build break in Terminal.

jeffdav:
There are three things that cause Terminal to not build on Windows:

1. `isprint()` returns an `Int32` which requires an explicit comparison. Easy fix: just compare against `0`.
2. `_getch()` returns an `Int32`, but `Unicode.Scalar()` takes a `UInt32`. So we must cast.
3. The entire implementation of `size` was Posix-specific.  Add a Windows implementation.

### DIFF
--- a/Sources/ConsoleKitTerminal/Terminal/Terminal.swift
+++ b/Sources/ConsoleKitTerminal/Terminal/Terminal.swift
@@ -1,5 +1,8 @@
 import Foundation
 import NIOConcurrencyHelpers
+#if os(Windows)
+import WinSDK
+#endif
 
 /// Generic console that uses a mixture of Swift standard
 /// library and Foundation code to fulfill protocol requirements.
@@ -72,8 +75,8 @@ public final class Terminal: Console, Sendable {
                 let c = _getch()
                 if c == 0x0d || c == 0x0a {
                     break
-                } else if isprint(c) {
-                    pass.append(Character(Unicode.Scalar(c)))
+                } else if isprint(c) != 0, let scalar = Unicode.Scalar(UInt32(c)) {
+                    pass.append(Character(scalar))
                 }
             }
 #endif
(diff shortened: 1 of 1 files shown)


jeffdav:
Note: To build on Windows, first one must update `swift-nio` to this revision or later:
https://github.com/apple/swift-nio/commit/8b66b22895115954dca2f1a6c5ab67172b6a6a1c

0xTim:
@jeffdav Could you enable Windows CI so we can track this properly? See https://github.com/vapor/jwt-kit/blob/main/.github/workflows/test.yml#L35

0xTim:
Yeah I'm not expecting it to pass but we can at least track the errors across the modules and make sure we don't add regressions in new code. It will be a while until NIO ships a windows compatible version

jeffdav:
@0xTim https://github.com/vapor/console-kit/pull/206
I'm not super familiar with your workflows but it seems straightforward enough.  LMK if this isn't what you meant.

BTW when I run `swift test` with these fixes + the `swift-nio` fixes everything does pass.  I.e. with this locally:
```
    ],
    dependencies: [
        .package(url: "https://github.com/apple/swift-log.git", from: "1.5.3"),
        .package(url: "https://github.com/apple/swift-nio.git", revision: "8b66b22895115954dca2f1a6c5ab67172b6a6a1c"),
    ],
```

It looks like `swift-nio` does releases pretty frequently, so hopefully won't have to wait too long for it!

0xTim:
Ah yeah, we're only using NIOConcurrency helpers, not the rest of NIO. Pity we had to pull that in but those fixes should land soon!

jeffdav:
SwiftNIO released 2.76.0 with the fixes in it.  I believe a rerun of the windows-unit workflow will pass now.

0xTim (approved):
Thanks!

PR #212, opened 2025-06-01
Remove Foundation

fpseverino:
Remove Foundation and use C libraries instead

### DIFF
--- a/Sources/ConsoleKit/Terminal/ANSI.swift
+++ b/Sources/ConsoleKit/Terminal/ANSI.swift
@@ -1,7 +1,15 @@
-import Foundation
-
-#if canImport(Android)
-import Android
+#if os(macOS)
+import Darwin.C
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif canImport(Bionic)
+import Bionic
+#elseif os(WASI)
+import WASILibc
+#elseif os(Windows)
+import CRT
 #endif
 
 /// Terminal ANSI commands
@@ -53,7 +61,6 @@ extension Terminal {
         // fdopen() on stdout is fast; also the returned file MUST NOT be fclose()d
         // This avoids concurrency complaints due to accessing global `stdout`.
         fflush(fdopen(STDOUT_FILENO, "w+"))
-
     }
 }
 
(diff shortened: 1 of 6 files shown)


MahdiBM:
I wonder if we really are no longer depending on the full Foundation, considering we don't have these flags enabled to make sure we don't have any "implicit" unintended imports:
```
        .enableUpcomingFeature("MemberImportVisibility"),
        .enableUpcomingFeature("InternalImportsByDefault"),
``` 
perhaps this should be for another PR? The flags _might_ require a decent amount of explicit imports to be added.

Sources/ConsoleKit/Terminal/ANSI.swift
```diff
@@ -1,7 +1,15 @@
-import Foundation
-
-#if canImport(Android)
+#if os(macOS)
```
MahdiBM, line 1: this should probably be `#if canImport(Darwin)`.
I'd try to refer to SwiftNIO in how to import these different modules. Maybe some stuff won't apply to this project, but i think most stuff do apply here to.

MahdiBM: example from NIO: https://github.com/apple/swift-nio/blob/37094b1b424d01ee9ad1b9188a35ab1204ef1db9/Sources/NIOFileSystem/Internal/SystemFileHandle.swift#L20

I'd just search e.g. `canImport(Darwin)` and try to work out if we should do what they've done in those places, which most of the times should be a yes.

Sources/ConsoleKit/Terminal/Console.swift
```diff
@@ -1,7 +1,15 @@
...
+#if os(macOS)
+import Darwin.C
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif canImport(Android)
 import Android
+#elseif os(WASI)
+import WASILibc
```
MahdiBM, line 10: Are you sure we _need_ these imports? Like if you remove them the app will no longer compile. I'm talking generally about all files, not this file specifically.

MahdiBM:
I'd wait for Tim/Gwynne review as well.

fpseverino:
> I'd wait for Tim/Gwynne review as well.

Whoops. 😅

PR #215, opened 2025-07-25
Fix activity indicator bug and update API

fpseverino:
- Fix bug of activity indicator that continued printing after the Task was cancelled
- `withActivityIndicator` now shows the "success" or "failure" state automatically based on whether the function throws an error or not
- Add a title to show next to `CustomActivity`
- Improve code coverage

### DIFF
--- a/Sources/ConsoleKit/Activity/CustomActivity.swift
+++ b/Sources/ConsoleKit/Activity/CustomActivity.swift
@@ -83,7 +83,7 @@ extension Console {
 
 /// An activity indicator with customizable frames and success and failure messages.
 ///
-/// See ``Console/customActivity(frames:success:failure:color:)`` to make one.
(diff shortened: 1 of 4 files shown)


Tests/ConsoleKitTests/ConsoleTests.swift
```diff
@@ -1,6 +1,7 @@
-import ConsoleKit
 import Testing
 
+@testable import ConsoleKit
```
0xTim, line 3: Do we need `@testable`? Ideally we'd only be testing the public API

fpseverino: It's just for covering some internal APIs and get an higher code coverage

fpseverino: I refactored the tests to keep the same code coverage without using `@testable`

PR #233, opened 2026-04-01
Cherry-pick the update for swift-log 1.11.0 from main

gwynne:
Title says it all 🙂. See #232 for detail.

### DIFF
--- a/Sources/ConsoleLogger/ConsoleLogger.swift
+++ b/Sources/ConsoleLogger/ConsoleLogger.swift
@@ -6,15 +6,16 @@ public import Configuration
 
 /// Outputs logs to console via a ``LoggerFragment`` pipeline.
 public struct ConsoleLogger<T: LoggerFragment>: LogHandler, Sendable {
+    /// The log handler's label.
     public let label: String
 
-    /// See `LogHandler.metadata`.
+    // See `LogHandler.metadata`.
     public var metadata: Logger.Metadata
 
-    /// See `LogHandler.metadataProvider`.
+    // See `LogHandler.metadataProvider`.
     public var metadataProvider: Logger.MetadataProvider?
 
-    /// See `LogHandler.logLevel`.
+    // See `LogHandler.logLevel`.
     public var logLevel: Logger.Level
 
     /// The ``LoggerFragment`` this logger outputs through.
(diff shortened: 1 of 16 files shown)


Sources/ConsoleLogger/ConsoleLogger.swift
```diff
@@ -6,15 +6,16 @@ public import Configuration
 
 /// Outputs logs to console via a ``LoggerFragment`` pipeline.
 public struct ConsoleLogger<T: LoggerFragment>: LogHandler, Sendable {
+    /// The log handler's label.
     public let label: String
 
-    /// See `LogHandler.metadata`.
```
fpseverino, line 11: Maybe they were updated to be DocC comment in the `consolekit-5` branch but were still normal comments in the main branch? See also the two other properties below

fpseverino (changes requested):
Thanks! Just a comment on that small DocC issue.

Is Yams broken on Musl?

I see you've also added some fixes for Windows, do you think Windows CI could work now?

Sources/ConsoleLogger/ConsoleLogger.swift
gwynne: No; they were originally DocC comments on the main branch as well, and the associated PR undid that. It's deliberate—for symbols inherited from a protocol, adding DocC comments to the conformance suppresses the existing docs, which is undesirable unless there's new information to provide—which in this case there isn't. This way, this particular symbol (for example) will get the docs from [Logging/LogHandler/metadata](https://swiftpackageindex.com/apple/swift-log/1.11.0/documentation/logging/loghandler/metadata), which is what we want.

gwynne:
Yes, the Windows CI now works on main, so it should work here as well. I'll add it to the workflow.

As for Yams, it seems like it is, yeah. Should probably check if it's being addressed upstream yet.

Sources/ConsoleLogger/ConsoleLogger.swift
fpseverino: Oh ok, got it! 👍🏻

fpseverino:
Seems like Swift Configuration doesn't compile on Windows 🤷‍♂️

Edit: but only on Swift 6.2(?)

fpseverino (approved):

PR #244, opened 2026-08-13
Modernise

fpseverino:
Add modernisation changes as per https://github.com/orgs/vapor/projects/17?pane=issue&itemId=227197885

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,9 +14,11 @@ jobs:
       contents: read
     uses: vapor/ci/.github/workflows/run-unit-tests.yml@main
     with:
+      with_release_mode_testing: true
       with_windows: true
       with_musl: true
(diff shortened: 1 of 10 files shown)


0xTim (commented):
It's a long shot, but can any of the unsafe functions be taken over by APIs in Swift System? The answer could well be no, but worth checking

fpseverino:
> It's a long shot, but can any of the unsafe functions be taken over by APIs in Swift System? The answer could well be no, but worth checking

At first glance, I think we can only use Swift System in a few lines of `linux_readpassphrase`, but I’m not sure if it’s worth the effort. Besides, that function is quite specific, so I’d prefer to change it as little as possible

0xTim (approved):
Pity about all the unsafe stuff but that's the deal with interacting with glibc!

PR #245, opened 2026-08-13
Include information from LogEvent.error in log output if present

### LINKED ISSUE #239: Support error parameter from swift-log
### Please summarize your feature request

Error information is not included in logs from `ConsoleLogger` when using the new overloads in `swift-log`.

### Describe the functionality you're seeking in more detail

Version `1.12.0` of `swift-log` added a new log methods that allows the caller to pass an `error` parameter that is propagated to the `LogHandler` (https://github.com/apple/swift-log/pull/425). It would be great if the `LogHandler` in this package (`ConsoleLogger`) used the value (if present) and included it in the final output. A simple serialization involved simply taking the string representation (`"\(error)"`)  and the type name (`String(reflecting: type(of: error))`), but of the implementation is free to choose the exact format.

If you agree with this suggestion, I'd be happy to file a PR with an implementation.

### Have you considered any alternatives?

_No response_

samuelmurray:
In `swift-log`, a user can attach an error object to a log entry, which is passed as-is to the `LogHandler` for it to serialise. `ConsoleLogger`, and the default fragment, is updated to use this property, serialising it as `[ Fully.Qualified.Type: string representation ]`.

Example:

```swift
do {
  try throwingFunc()
} catch {
  logger.warning("Something went wrong", error: error) // Outputs "[...] Something went wrong [ ThrownError: description ] [...]"
}
```

Fixes #239

### DIFF
--- a/Sources/ConsoleLogger/LogRecord.swift
+++ b/Sources/ConsoleLogger/LogRecord.swift
@@ -5,6 +5,7 @@ public struct LogRecord {
     public init(
         level: Logger.Level,
         message: Logger.Message,
+        error: (any Error)?,
         metadata: Logger.Metadata? = nil,
         source: String,
         file: String,
@@ -17,6 +18,7 @@ public struct LogRecord {
     ) {
         self.level = level
         self.message = message
+        self.error = error
         self.metadata = metadata
         self.source = source
         self.file = file
@@ -32,6 +34,8 @@ public struct LogRecord {
     public var level: Logger.Level
     /// The logged message
     public var message: Logger.Message
+    /// The error associated with the logged message
+    public var error: (any Error)?
     /// The metadata explicitly associated with the logged message
     public var metadata: Logger.Metadata?
     /// The source of the log message, usually the module name
--- a/Sources/ConsoleLogger/LoggerFragments/LoggerFragment.swift
+++ b/Sources/ConsoleLogger/LoggerFragments/LoggerFragment.swift
@@ -259,6 +259,21 @@ public struct MessageFragment: LoggerFragment {
     }
 }
 
+/// Writes the logger error to the output, and requests a separator for the next fragment only if the error was not nil
+public struct ErrorFragment: LoggerFragment {
+    public init() {}
+
+    public func hasContent(record: inout LogRecord) -> Bool {
+        record.error != nil
+    }
+
+    public func write(_ record: inout LogRecord, to output: inout FragmentOutput) {
+        guard let error = record.error else { return }
+        output += "[ \(String(reflecting: type(of: error))): \(error) ]"
+        output.needsSeparator = true
+    }
+}
+
 /// Writes the combined metadata to the output, and requests a separator for the next fragment only if the metadata was not empty.
 ///
 /// This fragment is considered to not have content if the metadata is empty.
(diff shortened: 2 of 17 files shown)


Sources/ConsoleLogger/LogRecord.swift
```diff
@@ -5,6 +5,7 @@ public struct LogRecord {
     public init(
         level: Logger.Level,
         message: Logger.Message,
+        error: (any Error)?,
```
samuelmurray, line 8: I'm assuming breaking changes are ok, since it's a beta release?

samuelmurray:
I didn't spend too much time thinking of a good output format for the error, so I'm happy to adjust if you have any suggestions. One alternative is to simply treat it as metadata, and use some sensible keys, eg. `error.type` and  `error.message`. This is how `StreamLogHandler` in `swift-log` does it (https://github.com/apple/swift-log/blob/main/Sources/Logging/Handlers/StreamLogHandler.swift#L197-L200), though I'm not suggesting Vapor has to do the same.

Tests/ConsoleLoggerTests/TestError.swift
```diff
@@ -0,0 +1,5 @@
+struct TestError: Error, CustomStringConvertible {
```
fpseverino: I would move this type inside the test case, so we don't pollute the namespace

fpseverino (approved):
LGTM, the placement of the `ErrorFragment` is correct IMO, I'm not too sure about the output format, but we can change that later without breaking API

Tests/ConsoleLoggerTests/TestError.swift
samuelmurray: Are you ok with having it in the test suite struct? Putting it in the test case func gives a strange type name: (eg `ConsoleLoggerTests.LoggerFragmentBuilderTests.(unknown context at $106764a38).(unknown context at $106764a44).TestError`)

fpseverino: Fine by me 👍🏻

0xTim (changes requested):
Error output should definitely go in `metadata`. Only static strings should be added to the actual message as those are what are queried so need to be the same

samuelmurray:
> Error output should definitely go in `metadata`. Only static strings should be added to the actual message as those are what are queried so need to be the same

@0xTim would you mind clarifying your suggestion? Do you mean we should skip adding a `ErrorFragment` struct, and instead modify `MetadataFragment` to also serialize the `error` parameter?

0xTim:
@samuelmurray yep exactly. Having a separate `ErrorFragment` complicates how we actually output information in the console, which then affects any tooling that relies on it. We want the error itself to be added to the metadata, which is already rendered, not as an additional fragment

Sources/ConsoleLogger/LoggerFragments/LoggerFragment.swift
```diff
@@ -259,18 +259,22 @@ public struct MessageFragment: LoggerFragment {
...
+        !record.allMetadata().isEmpty || record.error != nil
     }
 
     public func write(_ record: inout LogRecord, to output: inout FragmentOutput) {
-        let allMetadata = record.allMetadata()
+        var allMetadata = record.allMetadata()
+        if let error = record.error {
+            allMetadata["error.message"] = "\(error)"
+            allMetadata["error.type"] = "\(String(reflecting: type(of: error)))"
+        }
```
samuelmurray, line 277: Would you prefer to instead have this be part of `allMetadata()`?

samuelmurray:
@0xTim have you had time to look at my changes? Let me know if you have any other suggestions!

